PR #2, opened 2023-01-20
solution

Lun4trick:
[DEMO LINK](https://lun4trick.github.io/react_autocomplete/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,55 +1,77 @@
-import React from 'react';
+import React, { useCallback, useEffect, useState } from 'react';
+import { debounce } from 'lodash';
+import cn from 'classnames';
 import './App.scss';
 import { peopleFromServer } from './data/people';
+import { Person } from './types/Person';
+import { Autocomplete } from './components/Autocompete';
 
 export const App: React.FC = () => {
-  const { name, born, died } = peopleFromServer[0];
+  const [selectedPersonSlug, setSelectedPersonSlug] = useState('');
+  const [query, setQuery] = useState('');
+  const [debouncedQuery, setDebouncedQuery] = useState('');
+  const [isTyping, setIsTyping] = useState(false);
+  const selectedPerson: Person | undefined = peopleFromServer
+    .find(person => person.slug === selectedPersonSlug);
+
+  const applyQuery = useCallback(
+    debounce(setDebouncedQuery, 1000),
+    [],
+  );
+
+  const onSelect = (slug: string) => {
+    setSelectedPersonSlug(slug);
+  };
+
+  const applyTypingDelay = useCallback(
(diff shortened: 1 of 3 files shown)


src/components/Autocompete.tsx
```diff
@@ -0,0 +1,43 @@
+import React from 'react';
+import '../App.scss';
+import { Person } from '../types/Person';
+
+type Props = {
+  dropDownList: Person[]
+  onSelect: (slug: string) => void
+};
+
+export const Autocomplete: React.FC<Props> = ({ dropDownList, onSelect }) => {
```
nurdanokcu, line 10: image

I ran some tests,
Whenever anything changes in input field autocomplete component is being rendered. 
You can use Memo for that component

src/App.tsx
```diff
@@ -1,55 +1,77 @@
...
+    applyTypingDelay(false);
+  };
+
+  useEffect(() => {
+    setQuery('');
+    setDebouncedQuery('');
+    setIsTyping(false);
+  }, [selectedPerson]);
+
+  const dropDownList = peopleFromServer.filter(person => (
```
nurdanokcu, line 44: I couldn't find the implementation of the following requirement:

don't run filtering again if the text has not changed (pause in typing happened when the text was the same as before)

this can be achieved by using useEffect while filtering so it filters only when debounced query changes.

nurdanokcu (changes requested):
Nice implementation with debounce 
I added my feedback for optimization. 
Let me know what you think.

nurdanokcu (approved):

src/App.tsx
```diff
@@ -1,55 +1,79 @@
...
 
   return (
     <main className="section">
       <h1 className="title">
-        {`${name} (${born} = ${died})`}
+        {
+          selectedPerson
+            ? `${selectedPerson?.name} (${selectedPerson?.born} = ${selectedPerson?.died})`
+            : 'No selected person'
+        }
```
naziamoff, line 57: looks more ergonomic this way
```suggestion
        {selectedPerson
            ? `${selectedPerson?.name} (${selectedPerson?.born} = ${selectedPerson?.died})`
            : 'No selected person'}
```

src/components/Autocompete.tsx
```diff
@@ -0,0 +1,43 @@
...
+        dropDownList.length
+          ? dropDownList.map((person: Person) => (
+            <button
+              type="button"
+              className="dropdown-item button-normalize"
+              key={person.slug}
+              onClick={() => onSelect(person.slug)}
+            >
+              <p
+                className={
```
naziamoff, line 23: classnames lib

naziamoff (approved):
good, but make sure to take a look at the classnames library

PR #5, opened 2023-01-20
Solution

UmizDemud:
[DEMO LINK](https://UmizDemud.github.io/react-autocorrect/)

This was fun

### DIFF
--- a/src/components/Dropdown/Dropdown.tsx
+++ b/src/components/Dropdown/Dropdown.tsx
@@ -1,0 +1,209 @@
+import React, {
+  useCallback,
+  useEffect,
+  useRef,
+  useState,
+} from 'react';
+import { debounce } from 'lodash';
+import cn from 'classnames';
+import './index.css';
+
+type Props = {
+  placeholder: string;
+  words: string[];
+  delay: number;
+  save: (name: string) => void;
+};
+
+export const Dropdown: React.FC<Props> = ({
+  words,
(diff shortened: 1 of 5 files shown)


src/components/Dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,209 @@
...
+      setPending(true);
+      setSuggestionsDebounced();
+    }, [],
+  );
+
+  const handleKeyDown = useCallback(
+    (e: React.KeyboardEvent<HTMLInputElement>) => {
+      switch (e.key) {
+        case 'ArrowDown':
+          if (selectedSug < suggestions.length - 1) {
```
naziamoff, line 94: seems like a helper function would be good option here too

src/components/Dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,209 @@
...
+                save(inputRef.current.value);
+                inputRef.current.value = '';
+              }
+            } else {
+              inputRef.current.value = suggestions[selectedSug];
+              save(suggestions[selectedSug]);
+              assertVisible(-1);
+              setSelectedSug(-1);
+            }
+          }
```
naziamoff, line 126: needs some refactoring. Seems that you can create a couple of variables and make it inside one `if`

naziamoff (approved):
well done🔥 🔥

PR #19, opened 2023-05-04
Solve

Dynamizx:


src/App.tsx
```diff
@@ -1,54 +1,87 @@
...
+  useState, useRef, useMemo, useCallback, useEffect,
+} from 'react';
+import debounce from 'lodash.debounce';
+import { Person } from './types/Person';
 import { peopleFromServer } from './data/people';
 
-export const App: React.FC = () => {
-  const { name, born, died } = peopleFromServer[0];
+export const App = () => {
+  const [people] = useState<Person[]>(peopleFromServer);
```
anstsot, line 9: do you need the state for this variable?
```suggestion
  const people = peopleFromServer;
```

src/App.tsx
```diff
@@ -1,54 +1,87 @@
...
+
+  const handleInputChange = useCallback(
+    (event: React.ChangeEvent<HTMLInputElement>) => {
+      const search = event.target.value;
+
+      if (!search) {
+        setIsMenuOpen(false);
+      } else {
+        setIsMenuOpen(true);
+      }
```
anstsot, line 39: can be simplified
```suggestion
      setIsMenuOpen(!!search);
```

anstsot (changes requested):
Your demo link leads to the wrong task. could you fix it, please?

witflash (approved):
Good job! 👏

PR #22, opened 2023-05-16
Develop

RomeoKs:
[DEMO LINK](https://RomeoKs.github.io/react_autocomplete/)


src/components/DropdownMenu.tsx
```diff
@@ -0,0 +1,64 @@
...
+            <p
+              className="has-text-danger"
+            >
+              No matching suggestions
+            </p>
+          </div>
+        </div>
+      </div>
+    );
+  }
```
polosanya, line 34: Since you now check for searchResults.length === 0 in line 20, you can remove this condition from line 39. Just render the searchResults.map() inside the 'dropdown-content' div.

src/components/DropdownMenu.tsx
```diff
@@ -0,0 +1,64 @@
...
+
+export const DropdownMenu: React.FC<Props> = (
+  {
+    searchResults,
+    onSelect,
+    setIsDropdownActive,
+    activePersonSlug,
+  },
+) => {
+  if (!searchResults) {
```
polosanya, line 20: Change the condition to check for searchResults.length === 0 instead of !searchResults, as searchResults is of type Person[] | null.

src/App.tsx
```diff
@@ -1,56 +1,121 @@
...
-            className="input"
-          />
-        </div>
+  useEffect(() => {
+    if (isQueryCleared) {
+      setDelayedQuery('');
+      setSearchResults([]);
+      setIsQueryCleared(false);
+      setIsDropdownActive(false);
+    } else {
```
polosanya, line 31: Instead of using 'isQueryCleared', you can set 'query' to an empty string and handle this case directly in the useEffect watching 'query'. This way, you can remove 'isQueryCleared' state.

polosanya (changes requested):
Almost done!

src/App.tsx
RomeoKs: Is not only about watching the query, but is also about setting the searchResult to an empty array and the deactivating the dropdown.

I've modified the code to be used for example to search for cities in a shipping component (Ecommerce).

RomeoKs: @polosanya  Thank you for your input!

RomeoKs:
@polosanya  Thank you for your input!

PR #26, opened 2023-05-25
Solution

Irina-Kulish:
https://Irina-Kulish.github.io/react_autocomplete/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,53 +1,93 @@
-import React from 'react';
+import React, { useCallback, useMemo, useState } from 'react';
 import './App.scss';
+import cn from 'classnames';
 import { peopleFromServer } from './data/people';
+import { Person } from './types/Person';
 
-export const App: React.FC = () => {
-  const { name, born, died } = peopleFromServer[0];
+type Persons = {
+  peopleFromServer: Person[];
+};
+
+const debounce = (f:(...args: string[]) => void,
+  delay: number) => {
+  let timerId: NodeJS.Timeout;
+
+  return (...arg: string[]) => {
+    clearTimeout(timerId);
+    timerId = setTimeout(f, delay, ...arg);
+  };
+};
+
+export const App: React.FC<Persons> = () => {
+  const [query, setQuery] = useState('');
+  const [appliedQuery, setAppliedQuery] = useState('');
+  const [selectedPerson, setSelectedPerson] = useState<Person | null>(null);
+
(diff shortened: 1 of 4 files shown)


src/App.tsx
```diff
@@ -1,53 +1,89 @@
...
-
-            <div className="dropdown-item">
-              <p className="has-text-danger">Petronella de Decker</p>
-            </div>
+              {personFilter.map((person) => (
+                <button
+                  className="button"
+                  type="button"
+                  onClick={() => {
+                    setSelectedPerson(person); clearHandler();
```
mykhalenych, line 71: It's better to create a separate function (e.g. handlePersonSelect) for the onClick handler and call setSelectedPerson and clearHandler in that function.

src/App.tsx
```diff
@@ -1,53 +1,89 @@
...
+  return (...arg: string[]) => {
+    clearTimeout(timerId);
+    timerId = setTimeout(f, delay, ...arg);
+  };
+};
 
 export const App: React.FC = () => {
-  const { name, born, died } = peopleFromServer[0];
+  const [query, setQuery] = useState('');
+  const [appliedQuery, setAppliedQuery] = useState('');
```
mykhalenych, line 19: why do you need 2 states with query?
create only 1

src/App.tsx
```diff
@@ -1,53 +1,89 @@
...
+  const [selectedPerson, setSelectedPerson] = useState<Person | null>(null);
+
+  const applyQuery = useCallback(
```
mykhalenych, line 22: useCallback redundant here

src/App.tsx
TarasHoliuk: agree, but it's not critical

src/App.tsx
```diff
@@ -1,53 +1,86 @@
...
-            <div className="dropdown-item">
-              <p className="has-text-link">Pieter Bernard Haverbeke</p>
-            </div>
+              {personFilter.map((person) => (
+                <button
+                  className="button"
+                  type="button"
+                  key={person.slug}
+                  onClick={() => {
+                    setSelectedPerson(person); clearHandler();
```
TarasHoliuk, line 69:
```suggestion
                    setSelectedPerson(person); 
                    clearHandler();
```
at least, I'd write each one in the new line

TarasHoliuk (approved):
Left 1 comment (FYI) about code style - :heavy_check_mark:

PR #40, opened 2023-06-12
Autocomplete

Oleksii-Mishchenko:
- [DEMO LINK](https://Oleksii-Mishchenko.github.io/react_autocomplete/)


src/components/DropDownList/DropDownList.tsx
```diff
@@ -0,0 +1,44 @@
...
+export const DropDownList: React.FC<Props> = React.memo(
+  ({ suggestedPeople, onSelected }) => {
+    return (
+      <div className="dropdown-menu" role="menu">
+        <div className="dropdown-content">
+          {suggestedPeople.length
+            ? (
+              suggestedPeople.map(suggestedPerson => (
+                <a
+                  key={suggestedPerson.name}
```
polosanya, line 20: Using the person's name as a key may lead to problems if there are people with the same name. Consider using a unique identifier, like an ID, for the key prop.

src/App.tsx
```diff
@@ -1,14 +1,72 @@
...
+  useEffect,
+  useMemo, useState,
+} from 'react';
 import './App.scss';
 import { peopleFromServer } from './data/people';
+import { Person } from './types/Person';
+import { DropDownList } from './components/DropDownList';
+
+const debounce = (f: (value: string) => void, delay: number) => {
+  let timerId: NodeJS.Timeout;
```
polosanya, line 12: You can use a library like lodash for debounce function to ensure better performance and reliability.

polosanya (changes requested):
Almost done!
It would be great to show people initially, but it's up to you
image

polosanya (approved):

PR #43, opened 2023-06-15
Solution for react_autocomplete

petrovskyi-yevhen:
[DEMO LINK](https://petrovskyi-yevhen.github.io/react_autocomplete/)


Anastasiia-Svintsova (changes requested):
- Check the requirements. Hide suggestions on text change;

petrovskyi-yevhen:
> * Check the requirements. Hide suggestions on text change;

I try to add a state value 


``` const [CharAddedOrRemoved, setCharAddedOrRemoved] = useState(false); ``` 

and in **handleInput function** i set the state **CharAddedOrRemoved** to true. 

and change this 

``` <div className={CharAddedOrRemoved ? "dropdown is-active": "dropdown is-not-active"}></div>``` 

with adding to css dropdown is-not-active to hidden. 

But it doesn't work. How to do it?

File: src/components/People.tsx [hunk 1]
```diff
@@ -0,0 +1,69 @@
...
+            {filteredPeople.length > 0
+              && filteredPeople.map((person) => (
+                <button
+                  key={person.slug}
+                  type="button"
+                  className="dropdown-item person-input-item"
+                  onClick={() => {
+                    handleClick(
+                      `${person.name} (${person.born} - ${person.died})`,
+                    );
```
lerastarynets, line 46: Pls move template string into separate const and use here

src/components/People.tsx
```diff
@@ -0,0 +1,69 @@
...
+    setInput,
+    applyQueryWithDebounce,
+    filteredPeople,
+    handleClick,
+  } = useContext(PeopleContext);
+
+  const handleInput = (event: React.ChangeEvent<HTMLInputElement>) => {
+    const userInput = event.target.value;
+
+    applyQueryWithDebounce(userInput);
```
lerastarynets, line 17: You should pass the 'userInput' to the 'setInput' function before calling 'applyQueryWithDebounce' to ensure the input state is updated before debouncing.

src/components/People.tsx
```diff
@@ -0,0 +1,69 @@
...
+          className="input"
+          value={input}
+          onChange={handleInput}
+        />
+      </div>
+
+      {input && (
+        <div className="dropdown-menu" role="menu">
+          <div className="dropdown-content">
+            {filteredPeople.length > 0
```
lerastarynets, line 37: Could be just `filteredPeople.length` instead of `filteredPeople.length > 0`[as `0` is `false`](https://www.freecodecamp.org/news/falsy-values-in-javascript/)

lerastarynets (approved):
Great job overall! Left a few comments to make your code more advanced)

PR #44, opened 2023-06-17
react_autocomplete

YelyzavetaMolochynska:
[DEMO LINK](https://YelyzavetaMolochynska.github.io/react_autocomplete/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -93,7 +93,9 @@ export const App: React.FC<Props> = ({
 
             {error && (
               <div className="dropdown-item">
-                <p className="has-text-danger">No matching suggestions</p>
+                <p className="has-text-danger">
+                  No matching suggestions
+                </p>
               </div>
             )}
           </div>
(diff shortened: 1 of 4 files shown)


VitaliyBondarenko1982 (changes requested):
List of all people shouldn't be visible if query in input is empty
Looks like debounce not works
chrome-capture-2023-5-17

src/App.tsx
```diff
@@ -1,56 +1,104 @@
...
+    setCloseList(prevState => !prevState);
+  }
+
+  function handleInput(event: React.ChangeEvent<HTMLInputElement>) {
+    setQuery(event.target.value);
+    applyQuery(event.target.value);
+    setCloseList(false);
+  }
+
+  const error = visiblePeople.length === 0;
```
VitaliyBondarenko1982, line 53:
```suggestion
  const error = !!visiblePeople.length;
```
and maybe not need addittional variable error and just check it in jsx

VitaliyBondarenko1982 (approved):
Well done

PR #46, opened 2023-06-19
add autocomplete

DidenkoDenis306:
[DEMO LINK](https://DidenkoDenis306.github.io/react_autocomplete/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,14 +1,76 @@
-import React from 'react';
+import {
+  useMemo,
+  useState,
+  useCallback,
+  useEffect,
+} from 'react';
 import './App.scss';
 import { peopleFromServer } from './data/people';
+import { DropdownMenu } from './DropdownMenu';
+import { Person } from './types/Person';
 
-export const App: React.FC = () => {
-  const { name, born, died } = peopleFromServer[0];
+const debounce = <T extends string[]>(
+  f: (...args: T) => void,
+  delay: number,
+) => {
+  let timerId: NodeJS.Timeout;
+
+  return (...args: T) => {
+    clearTimeout(timerId);
+    timerId = setTimeout(f, delay, ...args);
+  };
+};
+
+export const App = () => {
+  const [query, setQuery] = useState<string>('');
+  const [appliedQuery, setAppliedQuery] = useState<string>('');
+  const [selectedPerson, setSelectedPerson] = useState<Person | null>(null);
+  const [showSuggestions, setShowSuggestions] = useState(false);
+
+  useEffect(() => {
+    if (!query.trim()) {
+      setShowSuggestions(false);
+    }
+  }, [query]);
+
+  const applyQuery = useCallback(
+    debounce(setAppliedQuery, 1000),
+    [],
+  );
+
+  const handleQueryOnChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+    const { value } = event.target;
+
+    setQuery(value);
+    applyQuery(value);
+    setShowSuggestions(false);
+  };
+
+  const onSelected = (person: Person) => {
+    setQuery(person.name);
+    setSelectedPerson(person);
+    setShowSuggestions(false);
+  };
+
+  const visiblePeople = useMemo(() => {
+    setShowSuggestions(true);
+
+    const normalizePersonName = (human: Person) => {
+      return human.name.toLowerCase().trim();
+    };
+
+    return peopleFromServer.filter(
+      person => normalizePersonName(person)
+        .includes(appliedQuery.toLowerCase()),
+    );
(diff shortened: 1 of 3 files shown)


src/App.tsx
```diff
@@ -1,14 +1,76 @@
...
+  const [query, setQuery] = useState<string>('');
+  const [appliedQuery, setAppliedQuery] = useState<string>('');
+  const [selectedPerson, setSelectedPerson] = useState<Person | null>(null);
+  const [showSuggestions, setShowSuggestions] = useState(false);
+
+  useEffect(() => {
+    if (!query.trim()) {
+      setShowSuggestions(false);
+    }
+  }, [query]);
```
dffUqp, line 34: can we replace this useEffect with a variable like?🤔

```jsx
const isSuggestionsVisible = !query.trim() && showSuggestions;
```

src/App.tsx
```diff
@@ -1,14 +1,76 @@
...
+
+  const applyQuery = useCallback(
+    debounce(setAppliedQuery, 1000),
+    [],
+  );
+
+  const handleQueryOnChange = (event: React.ChangeEvent<HTMLInputElement>) => {
```
dffUqp, line 41:
```suggestion
  const handleQueryChange = (event: React.ChangeEvent<HTMLInputElement>) => {
```

src/App.tsx
```diff
@@ -1,14 +1,76 @@
...
+    setShowSuggestions(true);
+
+    const normalizePersonName = (human: Person) => {
+      return human.name.toLowerCase().trim();
+    };
+
+    return peopleFromServer.filter(
+      person => normalizePersonName(person)
+        .includes(appliedQuery.toLowerCase()),
+    );
```
dffUqp, line 65: let's remove normalizePersonName function.

```suggestion
    return peopleFromServer.filter(
      person => {
        const normalizedPersonName = person.name.toLowerCase().trim()
      
        return normalizedPersonName.includes(appliedQuery.toLowerCase())
     });
```

dffUqp (changes requested):
LGTM:fire: but needs a few small fixes:upside_down_face:

src/App.tsx
DidenkoDenis306: Unfortunately, it does not work

src/DropdownMenu.tsx
```diff
@@ -0,0 +1,39 @@
...
+}
+
+export const DropdownMenu: React.FC<Props> = ({
+  people,
+  onSelected,
+}) => {
+  return (
+    <div className="dropdown-menu" role="menu">
+      <div className="dropdown-content">
+        {people.length > 0
```
dffUqp, line 16:
```suggestion
        {!people.length
```

src/DropdownMenu.tsx
```diff
@@ -0,0 +1,39 @@
...
+          ? (
+            people.map((person) => (
+              <div className="dropdown-item" key={person.slug}>
+                {/* eslint-disable-next-line jsx-a11y/no-noninteractive-element-interactions, jsx-a11y/click-events-have-key-events */}
+                <p
+                  className={person.sex === 'm'
+                    ? 'has-text-link'
+                    : 'has-text-danger'}
```
dffUqp, line 24: [classnames](https://www.npmjs.com/package/classnames  )

src/DropdownMenu.tsx
```diff
@@ -0,0 +1,39 @@
...
+  onSelected,
+}) => {
+  return (
+    <div className="dropdown-menu" role="menu">
+      <div className="dropdown-content">
+        {people.length > 0
+          ? (
+            people.map((person) => (
+              <div className="dropdown-item" key={person.slug}>
+                {/* eslint-disable-next-line jsx-a11y/no-noninteractive-element-interactions, jsx-a11y/click-events-have-key-events */}
```
dffUqp, line 20: let's try to fix this eslint issue. you give `onClick`  prop to paragraph, I think its not a good idea.

src/App.tsx
```diff
@@ -17,40 +77,17 @@ export const App: React.FC = () => {
             type="text"
             placeholder="Enter a part of the name"
             className="input"
+            value={query}
+            onChange={handleQueryOnChange}
```
dffUqp, line 81:
```suggestion
            onChange={handleQueryChange}
```

src/App.tsx
dffUqp: try this
`const isSuggestionsVisible = !!query.trim() && showSuggestions;`

src/DropdownMenu.tsx
DidenkoDenis306: Probably !!people.length? But elsint  says that's not the way to do it. So I think it's better to keep it that way.

PR #49, opened 2023-06-23
First solution

oleksandr-plias:
[DEMO LINK](https://oleksandr-plias.github.io/react_autocomplete/)

### DIFF
--- a/src/components/Autocomplete.tsx
+++ b/src/components/Autocomplete.tsx
@@ -1,0 +1,99 @@
+/* eslint-disable jsx-a11y/no-noninteractive-element-interactions */
+/* eslint-disable jsx-a11y/click-events-have-key-events */
+import React, {
+  useState,
+  useMemo,
+  useCallback,
+  Dispatch,
+} from 'react';
+import { peopleFromServer } from '../data/people';
+import { Person } from '../types/Person';
+
+type Props = {
+  delay: number;
+  onSelected: (username: Person) => void;
+};
+
+const normalize = (value: string) => value.toLowerCase();
+
+const debounce = (callback: Dispatch<string>, delay: number) => {
+  let timerId: number;
+
+  return (...args: string[]) => {
+    clearTimeout(timerId);
+
+    timerId = setTimeout(callback, delay, ...args);
+  };
+};
+
+export const Autocomplete: React.FC<Props> = ({ delay, onSelected }) => {
+  const [query, setQuery] = useState('');
+  const [appliedQuery, setAppliedQuery] = useState('');
+  const [listIsShown, setListIsShown] = useState('false');
+
+  const visibleUsers = useMemo(() => {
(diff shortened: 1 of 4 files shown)


src/components/Autocomplete.tsx
```diff
@@ -0,0 +1,99 @@
+/* eslint-disable jsx-a11y/no-noninteractive-element-interactions */
```
TarasHoliuk, line 1: try to get rid of it
some info: 
https://dev.to/shivampawar/error-non-interactive-elements-should-not-be-assigned-mouse-or-keyboard-event-jsx-a11yno-noninteractive-elements-interactions-2i1g

https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/main/docs/rules/no-noninteractive-element-interactions.md

src/components/Autocomplete.tsx
```diff
@@ -0,0 +1,99 @@
...
+    clearTimeout(timerId);
+
+    timerId = setTimeout(callback, delay, ...args);
+  };
+};
+
+export const Autocomplete: React.FC<Props> = ({ delay, onSelected }) => {
+  const [query, setQuery] = useState('');
+  const [appliedQuery, setAppliedQuery] = useState('');
+  const [listIsShown, setListIsShown] = useState('false');
```
TarasHoliuk, line 32:
```suggestion
  const [listIsShown, setListIsShown] = useState(false);
```
the reason list is shown everytime - use real boolean instead of string (fix everywhere)

TarasHoliuk:
The dropdown shouldn't be shown if nothing is typed and input is not in focus
image

TarasHoliuk (changes requested):
The logic looks very good :+1: 
But the one issue definitely should be fixed here (boolean usage)

src/components/Autocomplete.tsx
oleksandr-plias: Hi Taras. I have some problems. I fixed the logical problem you wrote about. Yes, I don’t have a list with users  at the beginning with an empty input now. But how do i get rid of the list after i have entered the text and cleared the input.  I tried to fix it but it doesn't work                                                                                                       
    const handleOnChange = (event: React.ChangeEvent<HTMLInputElement>) => {

    setListIsShown('false');
    setQuery(value);
    applyQuery(value);

    if (value.trim() === '') {
      setListIsShown(false);
    } else {
      applyQuery(value);
      setListIsShown(true);
    }
  };

oleksandr-plias: And I read the material on interactive elements, implemented it in my code, but for some reason es lint still gives an error                                                                                                                                                                                              
                <p
                    // role="button"
                    // tabIndex={0}
                    className="has-text-link"
                    onClick={() => handleOnClick(user)}
                  >

VitaliyBodnar (approved):
Well done 🔥

PR #52, opened 2023-06-25
Develop

serentinos:
[DEMO LINK](https://serentinos.github.io/react_autocomplete/)

Solution with UseContext. In [d607d7a](https://github.com/mate-academy/react_autocomplete/pull/52/commits/d607d7a19b1d1ea8c91a34f83cb4fa4881bf77b8) commit solution with only lifting states. Please leave some feedback about this usecase of useContext, thanks!

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,57 +1,22 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
 import { peopleFromServer } from './data/people';
+import { Person } from './types/Person';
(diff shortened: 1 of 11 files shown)


dffUqp:
it is really cool that you want to try use new things.:+1:

useContext we should in 2 cases
1. We can't use lifting state
2. Too much nesting of props

So in this situation useContex is unnecessary

dffUqp (approved):
Your solution with lifting states is approved:smiley:

PR #58, opened 2023-06-26
Solution

Shevtsooov:
[DEMO LINK](https://Shevtsooov.github.io/react_autocomplete/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,57 +1,63 @@
-import React from 'react';
-import './App.scss';
+import React, {
+  useCallback,
+  useMemo,
+  useState,
+} from 'react';
+
+import { debounce } from 'lodash';
 import { peopleFromServer } from './data/people';
+import { Person } from './types/Person';
+import { Dropdown } from './components/Dropdown';
+
+import './App.scss';
 
 export const App: React.FC = () => {
-  const { name, born, died } = peopleFromServer[0];
+  const [appliedQuery, setAppliedQuery] = useState('');
+  const [query, setQuery] = useState('');
+  const [people] = useState('');
+  const [selected, setSelected] = useState<Person | null>(null);
+
+  const applyQuery = useCallback(
+    debounce(
+      setAppliedQuery,
+      1000,
+    ),
+    [],
+  );
+
+  const selectPerson = useCallback((person) => {
+    setSelected(person);
+  }, [selected]);
+
+  const filteredPeople = useMemo(() => {
+    return peopleFromServer.filter(person => (
+      person.name.toLowerCase().includes(appliedQuery.toLowerCase())
+    ));
+  }, [people, appliedQuery]);
+
+  const handleReset = () => {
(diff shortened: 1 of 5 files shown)


src/App.tsx
```diff
@@ -1,57 +1,63 @@
...
-      </div>
+      <Dropdown
+        query={query}
+        setQuery={setQuery}
+        applyQuery={applyQuery}
+        appliedQuery={appliedQuery}
+        filteredPeople={filteredPeople}
+        onSelected={selectPerson}
+        handleReset={handleReset}
+      />
```
dinaminahi, line 60: Do we need all these props& or we can create some handlers?

dinaminahi (changes requested):
There is a bug we need to fix 

https://github.com/mate-academy/react_autocomplete/assets/46566190/9ceb190a-c8af-4635-bf9b-e0c4d74919bc

Shevtsooov:
@dinaminahi Hello, thanks for the detailed review.
I have fixed the bug and added handlers in my `Drodown` component, still I am not sure how to reduce the number of props.
Please suggest a solution.

src/components/Dropdown.tsx
```diff
@@ -0,0 +1,83 @@
...
+      </div>
+
+      {query && query === appliedQuery && (
+        <div className="dropdown-menu" role="menu">
+          <div className="dropdown-content">
+            {filteredPeople.length > 0 && (
+              filteredPeople.map(person => (
+                <div className="dropdown-item">
+                  {// eslint-disable-next-line jsx-a11y/click-events-have-key-events, jsx-a11y/no-noninteractive-element-interactions
+                    <p
```
dffUqp, line 65: let's fix this eslint issue:smiley: 

I think we can replace `<p>` with `<a>`

src/components/Dropdown.tsx
```diff
@@ -0,0 +1,83 @@
...
+
+  return (
+    <div className="dropdown is-active">
+      <div className="dropdown-trigger">
+        <input
+          type="text"
+          placeholder="Enter a part of the name"
+          className="input"
+          value={query}
+          onChange={(event) => handleSearch(event)}
```
dffUqp, line 45:
```suggestion
          onChange={handleSearch}
```

src/components/Dropdown.tsx
```diff
@@ -0,0 +1,83 @@
...
+            alt="reset"
+            onClick={handleReset}
+          />
+        )}
+      </div>
+
+      {query && query === appliedQuery && (
+        <div className="dropdown-menu" role="menu">
+          <div className="dropdown-content">
+            {filteredPeople.length > 0 && (
```
dffUqp, line 61: do we really need this check?

dffUqp (approved):
GJ✔️

about props i think in you `App.tsx` you need only `selectedPerson` so the rest of the variables you can move to `DropDown.tsx`

PR #61, opened 2023-06-26
add task solution

NickMitchell61:
[DEMO LINK](https://NickMitchell61.github.io/react_autocomplete/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,55 +1,104 @@
-import React from 'react';
+import {
+  ChangeEvent,
+  FC,
+  MouseEvent,
+  useCallback,
+  useEffect,
+  useState,
+} from 'react';
+import cn from 'classnames';
 import './App.scss';
 import { peopleFromServer } from './data/people';
+import { Person } from './types/Person';
+import { DropDownContent } from './components/DropDownContent/DropDownContent';
 
-export const App: React.FC = () => {
-  const { name, born, died } = peopleFromServer[0];
+const debounce = <T extends (...args: never[]) => void>(
+  func: T,
+  delay: number,
+) => {
+  let timerId: NodeJS.Timeout;
+
+  return (...args: Parameters<T>): void => {
+    clearTimeout(timerId);
+    timerId = setTimeout(func, delay, ...args);
+  };
+};
+
+export const App: FC = () => {
+  const [peoples, setPeoples] = useState<Person[]>([]);
+  const [query, setQuery] = useState('');
+  const [isSelectedQuery, setIsSelectedQuery] = useState(false);
+  const [selectedSlug, setSelectedSlug] = useState('');
+
+  useEffect(() => {
+    setPeoples(peopleFromServer);
+  }, []);
+
+  const queryWithDelay = useCallback(debounce(setIsSelectedQuery, 300), []);
+
+  const changeQuery = (event: ChangeEvent<HTMLInputElement>) => {
+    const { value } = event.target;
+
+    setQuery(value);
+
+    setQuery(value);
+    if (!value) {
+      queryWithDelay(false);
+    } else {
+      queryWithDelay(true);
+    }
+  };
+
+  const onSelectPerson = (
+    event: MouseEvent,
+    person: Person,
+  ) => {
+    event.preventDefault();
+
+    setSelectedSlug(person.slug);
+    setQuery(person.name);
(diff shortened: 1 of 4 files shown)


src/App.tsx
```diff
@@ -1,55 +1,104 @@
...
+
+export const App: FC = () => {
+  const [peoples, setPeoples] = useState<Person[]>([]);
+  const [query, setQuery] = useState('');
+  const [isSelectedQuery, setIsSelectedQuery] = useState(false);
+  const [selectedSlug, setSelectedSlug] = useState('');
+
+  useEffect(() => {
+    setPeoples(peopleFromServer);
+  }, []);
```
33Orange, line 35:
```suggestion
```

src/App.tsx
```diff
@@ -1,55 +1,104 @@
...
+  return (...args: Parameters<T>): void => {
+    clearTimeout(timerId);
+    timerId = setTimeout(func, delay, ...args);
+  };
+};
+
+export const App: FC = () => {
+  const [peoples, setPeoples] = useState<Person[]>([]);
+  const [query, setQuery] = useState('');
+  const [isSelectedQuery, setIsSelectedQuery] = useState(false);
```
33Orange, line 30: Let's create one more `useState` with `appliedQuery` and rename boolean `useState`.
Boolean `useState` we will use to render our list.
`appliedQuery` we will use for creating our `visiblePeople`

```suggestion
  const [query, setQuery] = useState('');
  const [appliedQuery, setAppliedQuery] = useState('');
  const [isSuggestionsActive, setIsSuggestionsActive] = useState(false);
```

src/App.tsx
```diff
@@ -1,55 +1,104 @@
...
+  const [selectedSlug, setSelectedSlug] = useState('');
+
+  useEffect(() => {
+    setPeoples(peopleFromServer);
+  }, []);
+
+  const queryWithDelay = useCallback(debounce(setIsSelectedQuery, 300), []);
```
33Orange, line 37: Here we will be creating function wich will set new state for `appliedQuery` with delay

```suggestion
  const applyQuery = useCallback(debounce(setAppliedQuery, 1000), []);
```

src/App.tsx
```diff
@@ -1,55 +1,104 @@
...
+    const { value } = event.target;
+
+    setQuery(value);
+
+    setQuery(value);
+    if (!value) {
+      queryWithDelay(false);
+    } else {
+      queryWithDelay(true);
+    }
```
33Orange, line 49:
```suggestion
    setQuery(value);
    applyQuery(value);
    setIsSuggestionsActive(false);
```

src/App.tsx
```diff
@@ -1,55 +1,104 @@
...
+    event.preventDefault();
+
+    setSelectedSlug(person.slug);
+    setQuery(person.name);
+    setIsSelectedQuery(false);
+  };
+
+  const visiblePeoples = peoples.filter(person => (
+    person.name.toLowerCase().trim().includes(query.toLowerCase())
+  ));
```
33Orange, line 65: Wrap it to `useMemo` and change flag isSuggestionsActive when you create new `visiblePeople`
```suggestion
  const visiblePeople = peoples.filter(person => (
    person.name.toLowerCase().trim().includes(query.toLowerCase())
  ));
```

src/App.tsx
```diff
@@ -1,55 +1,104 @@
...
+
+  const getPeopleBySlug = (slug: string) => {
+    return peoples.find(person => person.slug === slug);
+  };
+
+  const selectedPerson = getPeopleBySlug(selectedSlug);
+  const { name, born, died } = selectedPerson ?? {};
```
33Orange, line 72: It looks like a crutch

```suggestion
  const { name, born, died } = selectedPerson ?? {};
```

src/App.tsx
```diff
@@ -1,55 +1,104 @@
...
-
-            <div className="dropdown-item">
-              <p className="has-text-danger">Elisabeth Hercke</p>
-            </div>
-          </div>
+          <DropDownContent
+            visiblePeoples={visiblePeoples}
+            selectedSlug={selectedSlug}
+            onSelectPerson={onSelectPerson}
+          />
```
33Orange, line 101: We should render this component only when `isSuggestionsActive === true`

33Orange (changes requested):
Almost done

A few moments to fix

PR #63, opened 2023-06-27
solution

Yuliya1907:
[DEMO LINK](https://Yuliya1907.github.io/react_autocomplete/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,14 +1,51 @@
-import React from 'react';
+import React, { useCallback, useMemo, useState } from 'react';
+import { debounce } from './utils';
 import './App.scss';
 import { peopleFromServer } from './data/people';
+import { DropList } from './components/DropList';
+import { Person } from './types/Person';
 
 export const App: React.FC = () => {
-  const { name, born, died } = peopleFromServer[0];
+  const [query, setQuery] = useState('');
+  const [querySearch, setQuerysearch] = useState('');
+  const [selectedPerson, setSelectedPerson] = useState<Person | null>(null);
+
+  const applyQuery = useCallback(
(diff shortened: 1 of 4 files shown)


src/App.tsx
```diff
@@ -1,14 +1,51 @@
...
+        const normalizedQuery = querySearch.toLowerCase();
+
+        return peopleFromServer.filter(
+          person => person.name.toLowerCase().includes(normalizedQuery),
+        );
+      }
+
+      return [];
+    }, [querySearch],
+  );
```
33Orange, line 35:
```suggestion
  const visiblePeople = useMemo(() => {
      if (!querySearch) {
        return [];
      }

      const normalizedQuery = querySearch.toLowerCase();

      return peopleFromServer.filter(
          person => person.name.toLowerCase().includes(normalizedQuery),
      );
    }, [querySearch],
  );
```

src/App.tsx
```diff
@@ -17,40 +54,17 @@ export const App: React.FC = () => {
...
-              <p className="has-text-danger">Elisabeth Hercke</p>
-            </div>
-          </div>
-        </div>
+        {querySearch && (
+          <DropList
+            onSelected={handlePerson}
+            visiblePeople={visiblePeople}
+          />
+        )}
```
33Orange, line 67: The dropdown should hide when I enter a new letter to input.
Let's add a new boolean state to handle this.
And change it when you: 
 - change input
 - choose suggestion
 - change it to show when we render a new array of visible people

33Orange (changes requested):
Almost done

33Orange (approved):
Well done

PR #67, opened 2023-06-27
1st version

Grechana:
https://Grechana.github.io/react_autocomplete/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,54 +1,89 @@
-import React from 'react';
+import React, { useCallback, useMemo, useState } from 'react';
 import './App.scss';
 import { peopleFromServer } from './data/people';
+import { Person } from './types/Person';
+
+const debounce = (func:React.Dispatch<React.SetStateAction<string>>,
+  delay:number) => {
+  let timerId: number;
+
+  return (...args:string[]) => {
+    clearTimeout(timerId);
+    timerId = setTimeout(func, delay, ...args);
+  };
+};
 
 export const App: React.FC = () => {
-  const { name, born, died } = peopleFromServer[0];
+  const [query, setQuery] = useState('');
+  const [appliedQuery, setAppliedQuery] = useState('');
+  const [selected, setSelected] = useState<Person>();
+
+  const applyQuery = useCallback(
+    debounce(setAppliedQuery, 1000),
+    [],
+  );
+
+  const visiblePeople = useMemo(() => {
+    return peopleFromServer.filter(
+      person => person.name.toLowerCase().includes(query.toLowerCase()),
+    );
+  }, [peopleFromServer, appliedQuery]);
 
   return (
     <main className="section">
       <h1 className="title">
-        {`${name} (${born} = ${died})`}
+        {selected
+          ? `${selected.name} (${selected.born} = ${selected.died})`
+          : 'No selected person'}
       </h1>
 
-      <div className="dropdown is-active">
+      <div
+        className={appliedQuery ? 'dropdown is-active' : 'dropdown'}
+      >
         <div className="dropdown-trigger">
           <input
             type="text"
             placeholder="Enter a part of the name"
             className="input"
+            value={query}
+            onChange={event => {
+              setQuery(event.target.value);
+              applyQuery(event.target.value);
+            }}
           />
         </div>
 
         <div className="dropdown-menu" role="menu">
           <div className="dropdown-content">
-            <div className="dropdown-item">
-              <p className="has-text-link">Pieter Haverbeke</p>
-            </div>
-
-            <div className="dropdown-item">
-              <p className="has-text-link">Pieter Bernard Haverbeke</p>
-            </div>
-
-            <div className="dropdown-item">
-              <p className="has-text-link">Pieter Antone Haverbeke</p>
-            </div>
-
-            <div className="dropdown-item">
-              <p className="has-text-danger">Elisabeth Haverbeke</p>
-            </div>
-
-            <div className="dropdown-item">
-              <p className="has-text-link">Pieter de Decker</p>
-            </div>
-
-            <div className="dropdown-item">
-              <p className="has-text-danger">Petronella de Decker</p>
-            </div>
-
-            <div className="dropdown-item">
-              <p className="has-text-danger">Elisabeth Hercke</p>
-            </div>
+            {visiblePeople.length >= 1
+              ? visiblePeople.map(person => (
+                <div
+                  key={person.name}
+                  className="dropdown-item"
+                >
+                  <a
+                    href="/"
+                    onClick={(event) => {
+                      event.preventDefault();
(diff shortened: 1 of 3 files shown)


src/App.tsx
```diff
@@ -1,54 +1,89 @@
...
-            </div>
+            {visiblePeople.length >= 1
+              ? visiblePeople.map(person => (
+                <div
+                  key={person.name}
+                  className="dropdown-item"
+                >
+                  <a
+                    href="/"
+                    onClick={(event) => {
```
lerastarynets, line 66: Using an anchor tag for non-navigation purpose is not a good practice. Consider using a button or another interactive element instead.

src/App.tsx
```diff
@@ -1,54 +1,89 @@
...
+  const applyQuery = useCallback(
+    debounce(setAppliedQuery, 1000),
+    [],
+  );
+
+  const visiblePeople = useMemo(() => {
+    return peopleFromServer.filter(
+      person => person.name.toLowerCase().includes(query.toLowerCase()),
+    );
+  }, [peopleFromServer, appliedQuery]);
```
lerastarynets, line 30: The dependency array for useMemo should include 'query', not 'appliedQuery' or 'peopleFromServer'. 'peopleFromServer' is not a state or prop and 'appliedQuery' is not used in the memoization function. The function is using 'query' variable which is not included in the dependency array.

src/App.tsx
```diff
@@ -1,54 +1,89 @@
...
+      >
         <div className="dropdown-trigger">
           <input
             type="text"
             placeholder="Enter a part of the name"
             className="input"
+            value={query}
+            onChange={event => {
+              setQuery(event.target.value);
+              applyQuery(event.target.value);
```
lerastarynets, line 51: Pls move this function to a separate const and use as a callback here

src/App.tsx
```diff
@@ -1,54 +1,89 @@
...
+                  key={person.name}
+                  className="dropdown-item"
+                >
+                  <a
+                    href="/"
+                    onClick={(event) => {
+                      event.preventDefault();
+                      setSelected(person);
+                      setAppliedQuery('');
+                    }}
```
lerastarynets, line 70: Same here move this function to a separate const

src/App.tsx
```diff
@@ -1,54 +1,89 @@
...
-            </div>
-
-            <div className="dropdown-item">
-              <p className="has-text-danger">Petronella de Decker</p>
-            </div>
-
-            <div className="dropdown-item">
-              <p className="has-text-danger">Elisabeth Hercke</p>
-            </div>
+            {visiblePeople.length >= 1
```
lerastarynets, line 58: Pls [see the doc](https://developer.mozilla.org/en-US/docs/Glossary/Falsy)
```suggestion
            {visiblePeople.length
```

lerastarynets (changes requested):
Nice start! Let's fix and improve your work a little bit)

polosanya (changes requested):
1. There is no inputted text, but the person is shown
image
2. There should be `-` instead of `=`
image
image
3. It would be great to show all people in the list initially
image

src/App.tsx
```diff
@@ -1,54 +1,107 @@
...
+    );
+  }, [appliedQuery]);
+
+  const changeInInput = (event: React.ChangeEvent<HTMLInputElement>) => {
+    setQuery(event.target.value);
+    applyQuery(event.target.value);
+    setshowPeople(true);
+  };
+
+  const selectPerson = (event: React.MouseEvent<HTMLButtonElement, MouseEvent>,
```
sTorba24, line 39:
```suggestion
  const handleSelectPerson = (event: React.MouseEvent<HTMLButtonElement, MouseEvent>,
```

sTorba24 (changes requested):
Good work! Lets improve it


Shouldn't be visible before i type
Simage

src/App.tsx
Grechana: противоречит, прошлому коментарию : It would be great to show all people in the list initially
image

Grechana: коментарий "It would be great to show all people in the list initially" противоречит "Shouldn't be visible before i type"

src/App.tsx
```diff
@@ -1,54 +1,108 @@
-import React from 'react';
+import React, { useCallback, useMemo, useState } from 'react';
 import './App.scss';
 import { peopleFromServer } from './data/people';
+import { Person } from './types/Person';
+
+const debounce = (func:React.Dispatch<React.SetStateAction<string>>,
+  delay:number) => {
```
polosanya, line 7:
```suggestion
const debounce = (
  func: React.Dispatch<React.SetStateAction<string>>,
  delay: number,
) => {
```
I think it's better to move this function into a separate file and import here, so you can reuse it later in other components

src/App.tsx
```diff
@@ -1,54 +1,108 @@
...
+    clearTimeout(timerId);
+    timerId = setTimeout(func, delay, ...args);
+  };
+};
 
 export const App: React.FC = () => {
-  const { name, born, died } = peopleFromServer[0];
+  const [query, setQuery] = useState('');
+  const [appliedQuery, setAppliedQuery] = useState('');
+  const [selected, setSelected] = useState<Person>();
```
polosanya, line 19:
```suggestion
  const [selectedPerson, setSelectedPerson] = useState<Person>();
```
It would be great to specify what exactly is selected

src/App.tsx
```diff
@@ -1,54 +1,108 @@
...
+  const [showPeople, setshowPeople] = useState(false);
```
polosanya, line 20:
```suggestion
  const [isDropDownActive, setIsDropDownActive] = useState(false);
```
'showPeople' sounds like a function name. Let's name the boolean variable according to these rules, so it's clear that it's a boolean
image

https://dev.to/michi/tips-on-naming-boolean-variables-cleaner-code-35ig

src/App.tsx
```diff
@@ -1,54 +1,108 @@
...
+
+  const handleInputChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+    setQuery(event.target.value);
+    applyQuery(event.target.value);
+    setshowPeople(true);
+  };
+
+  const handleSelectPerson
+  = (event: React.MouseEvent<HTMLButtonElement, MouseEvent>,
+    person: Person) => {
```
polosanya, line 41:
```suggestion
  const handleSelectPerson = (
    event: React.MouseEvent<HTMLButtonElement, MouseEvent>,
    person: Person,
  ) => {
```
image

polosanya (changes requested):
Almost perfect! Let's polish it up 🚀 
1. It would be great for option to be clickable for the whole width
2. Let's close the dropdown when the input is not focused (onBlur)
focus

agniya-i (approved):

PR #70, opened 2023-06-28
solution

vitvadym:
[DEMO LINK](https://vitvadym.github.io/react_autocomplete/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -38,6 +38,10 @@ export const App: React.FC = () => {
     applyQuery(event.target.value);
     setSearchQuery(event.target.value);
     setIsSuggestions(event.target.value !== '');
+
+    if (isSuggestions) {
+      setSelectedPerson(null);
+    }
   };
 
   const handlClearButton = () => {
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,56 +1,85 @@
-import React from 'react';
 import './App.scss';
+import React, { useCallback, useMemo, useState } from 'react';
+import cn from 'classnames';
+import { Person } from './types/Person';
 import { peopleFromServer } from './data/people';
+import { Dropdown } from './components/DropdownList';
+import { debounce } from './debounce';
 
 export const App: React.FC = () => {
-  const { name, born, died } = peopleFromServer[0];
+  const [people] = useState(peopleFromServer);
+  const [searchQuery, setSearchQuery] = useState('');
+  const [appliedQuery, setAppliedQuery] = useState('');
+  const [isSuggestions, setIsSuggestions] = useState(false);
+  const [selectedPerson, setSelectedPerson] = useState<Person | null>();
+
+  const applyQuery = useCallback(
+    debounce(setAppliedQuery, 1000),
+    [],
+  );
+
+  const { name, born, died } = selectedPerson ?? {};
+
+  const preparedSearchQuery = searchQuery.toLowerCase().trim();
+
+  const visiblePeople = useMemo(() => {
+    return people.filter(
+      person => person.name.toLowerCase().includes(preparedSearchQuery),
+    );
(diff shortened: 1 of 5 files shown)


src/App.tsx
```diff
@@ -1,56 +1,89 @@
...
+import React, { useCallback, useMemo, useState } from 'react';
+import cn from 'classnames';
+import { Person } from './types/Person';
 import { peopleFromServer } from './data/people';
+import { Dropdown } from './components/DropdownList';
+import { debounce } from './debounce';
 
 export const App: React.FC = () => {
-  const { name, born, died } = peopleFromServer[0];
+  const [people] = useState(peopleFromServer);
```
naziamoff, line 10: what's the purpose for this usestate?

src/App.tsx
```diff
@@ -1,56 +1,89 @@
...
+
+  const { name, born, died } = selectedPerson ?? {};
+
+  const preparedSearchQuery = searchQuery.toLowerCase().trim();
+
+  const visiblePeople = useMemo(() => {
+    return people.filter(
+      person => person.name.toLowerCase().includes(preparedSearchQuery),
+    );
+  }, [appliedQuery, people]);
```
naziamoff, line 29: if you don't have any actions before the return statement - you can return right away, using this syntax:  `=> (...here's what you return)`

src/App.tsx
```diff
@@ -1,56 +1,89 @@
...
+    if (isSuggestions) {
+      setSelectedPerson(null);
+    }
+  };
+
+  const handlClearButton = () => {
+    setSearchQuery('');
+    setIsSuggestions(false);
+    setSelectedPerson(null);
+  };
```
naziamoff, line 51: why not just clear?

src/components/DropdownList.tsx
```diff
@@ -0,0 +1,42 @@
+/* eslint-disable jsx-a11y/no-noninteractive-element-interactions */
+/* eslint-disable jsx-a11y/click-events-have-key-events */
```
naziamoff, line 2: bad idea, better to leave comments right before you gonna use them. But what's even better - not to disable linting rules, just fix them instead

src/components/DropdownList.tsx
```diff
@@ -0,0 +1,42 @@
...
+interface Props {
+  people: Person[];
+  selectPerson: (person: Person) => void;
+}
+
+export const Dropdown: FC<Props> = ({
+  selectPerson,
+  people,
+}) => {
+  return (
```
naziamoff, line 16: you can return right away here

src/components/DropdownList.tsx
```diff
@@ -0,0 +1,42 @@
...
+    <div className="dropdown-menu" role="menu">
+      <div className="dropdown-content">
+        {people.length !== 0
```
naziamoff, line 19: probably you meant this one?
```suggestion
        {people.length > 0
```

naziamoff (changes requested):
looks good 👍 
A couple of minor suggestions and we're good

src/App.tsx
```diff
@@ -1,56 +1,89 @@
...
-
-            <div className="dropdown-item">
-              <p className="has-text-link">Pieter de Decker</p>
-            </div>
-
-            <div className="dropdown-item">
-              <p className="has-text-danger">Petronella de Decker</p>
-            </div>
+          <span className="icon">
+            {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
```
roman-mirzoian, line 74: bad idea, better to leave comments right before you gonna use them. But what's even better - not to disable linting rules, just fix them instead

roman-mirzoian (changes requested):
The application works well and is well written, but I see that some comments remain unanswered. Please correct them or discuss their feasibility with the author of the comments.

PR #72, opened 2023-06-30
autocomplete was added

Pavlo-Petrashevskyi:
[DEMO LINK](https://Pavlo-Petrashevskyi.github.io/react_autocomplete/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,54 +1,115 @@
-import React from 'react';
+import React, { useState, useMemo, useCallback } from 'react';
 import './App.scss';
+import classNames from 'classnames';
 import { peopleFromServer } from './data/people';
+import { Person } from './types/Person';
+
+const debounce = (
+  func: React.Dispatch<React.SetStateAction<string>>
+  | React.Dispatch<React.SetStateAction<boolean>>,
+  delay: number,
+) => {
+  let timer: number;
+
+  return (...args: any) => {
+    clearTimeout(timer);
+    timer = setTimeout(func, delay, ...args);
+  };
+};
 
 export const App: React.FC = () => {
-  const { name, born, died } = peopleFromServer[0];
+  const [selectedPerson, setSelectedPerson] = useState<Person>();
+  const [appliedQuery, setAppliedQuery] = useState('');
+  const [query, setQuery] = useState('');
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -1,54 +1,115 @@
...
+import { Person } from './types/Person';
+
+const debounce = (
+  func: React.Dispatch<React.SetStateAction<string>>
+  | React.Dispatch<React.SetStateAction<boolean>>,
+  delay: number,
+) => {
+  let timer: number;
+
+  return (...args: any) => {
```
loralevitska, line 14: Don't use any for types. Add the correct type here to fix lint warning (theoretically, the user should enter a string type there)
image

src/App.tsx
```diff
@@ -1,54 +1,115 @@
...
+              ? (visiblePeople.map(person => (
+                <div
+                  className="dropdown-item"
+                  key={person.name}
+                >
+                  <a
+                    href="http//"
+                    className={person.sex === 'm'
+                      ? 'has-text-link'
+                      : 'has-text-danger'}
```
loralevitska, line 98: Use [classnames](https://www.npmjs.com/package/classnames) library to add classes conditionally

src/App.tsx
```diff
@@ -1,54 +1,115 @@
...
+                    onClick={(e) => {
+                      e.preventDefault();
+                      handleSelected(person);
+                    }}
```
loralevitska, line 102: Move to separate function  (transfer the function to a variable, and here only write the name of the function)

loralevitska (changes requested):
Almost done) Let's improve

- this menu should be open when the input field has focus and now it is closed when the page is reloaded, please fix this
image

src/App.tsx
Pavlo-Petrashevskyi: done

src/App.tsx
Pavlo-Petrashevskyi: done

src/App.tsx
Pavlo-Petrashevskyi: done

loralevitska (approved):
Good job!

PR #85, opened 2023-07-08
React_autocomplete

ArtemKravcov85:
[DEMO LINK](https://ArtemKravcov85.github.io/react_autocomplete/)


src/components/DropDownMenu/DropDownMenu.tsx
```diff
@@ -0,0 +1,62 @@
...
+    person: Person,
+    status: boolean
+  ) => void;
+};
+
+export const DropDownMenu:React.FC<Props> = memo(({
+  visiblePeople,
+  onSelect,
+}) => {
+  const onSelected = (
```
lerastarynets, line 18: The 'onSelected' function is duplicating the functionality of the 'onSelect' prop. You could pass the 'onSelect' function directly to the onClick handler, provided the event signatures match.

src/App.tsx
```diff
@@ -1,56 +1,75 @@
...
+  const [selectedPerson, setSelectedPerson] = useState<Person | null>(null);
+  const [isSugActive, setIsSugActive] = useState(false);
+
+  const visiblePeople = useMemo(() => {
+    const normalizedQuery = appliedQuery.toLowerCase().trim();
+
+    setIsSugActive(!!normalizedQuery);
+
+    if (!normalizedQuery) {
+      setSelectedPerson(null);
```
lerastarynets, line 21: Avoid updating state within useMemo. useMemo should be a pure function without side effects. You can update 'selectedPerson' state based on 'appliedQuery' in useEffect instead.

src/App.tsx
```diff
@@ -1,56 +1,75 @@
...
-  const { name, born, died } = peopleFromServer[0];
+  const [query, setQuery] = useState('');
+  const [appliedQuery, setAppliedQuery] = useState('');
+  const [selectedPerson, setSelectedPerson] = useState<Person | null>(null);
+  const [isSugActive, setIsSugActive] = useState(false);
+
+  const visiblePeople = useMemo(() => {
+    const normalizedQuery = appliedQuery.toLowerCase().trim();
+
+    setIsSugActive(!!normalizedQuery);
```
lerastarynets, line 18: Avoid updating state within useMemo. useMemo should be a pure function without side effects. You can update 'isSugActive' state based on 'appliedQuery' in useEffect instead.

src/App.tsx
```diff
@@ -1,56 +1,75 @@
...
+  );
+
+  const handleSelect = useCallback((
+    event: React.MouseEvent<HTMLAnchorElement, MouseEvent>,
+    person: Person,
+    status: boolean,
+  ) => {
+    event.preventDefault();
+    setSelectedPerson(person);
+    setQuery(person.name);
```
lerastarynets, line 41: You should clear appliedQuery when a person is selected. This will ensure that when the user starts typing again, the autocomplete will consider the new input, not the previously selected person's name.

src/App.tsx
```diff
@@ -1,56 +1,75 @@
...
       <h1 className="title">
-        {`${name} (${born} = ${died})`}
+        {selectedPerson
+          ? `${selectedPerson?.name} (${selectedPerson?.born} - ${selectedPerson?.died})`
+          : 'No selected person'}
       </h1>
 
-      <div className="dropdown is-active">
+      <div className={cn('dropdown',
+        { 'is-active': isSugActive === true })}
```
lerastarynets, line 61:
```suggestion
        { 'is-active': isSugActive })}
```

lerastarynets (changes requested):
I see you did a hard work) Let's improve it a little bit

polosanya (approved):

PR #86, opened 2023-07-08
Develop

adv555:
**[DEMO LINK](https://adv555.github.io/react_autocomplete/)**

### DIFF
--- a/src/components/Dropdown/Dropdown.tsx
+++ b/src/components/Dropdown/Dropdown.tsx
@@ -1,0 +1,81 @@
+import React from 'react';
+import './Dropdown.scss';
+import cn from 'classnames';
+import { Person } from '../../types/Person';
+import { DropdownItem } from '../DropdownItem';
+
+type Props = {
+  people: Person[];
+  query: string;
+  debouncedQuery: string;
+  onQueryChange: (newQuery: string) => void;
+  onResetQuery: () => void;
+  onPersonClick: (person: Person) => void;
+};
+
+export const Dropdown: React.FC<Props> = (
+  {
+    people,
+    query,
+    debouncedQuery,
+    onQueryChange,
+    onResetQuery,
+    onPersonClick,
+  },
+) => {
+  return (
+    <div className={
+      cn(
+        'dropdown',
+        { 'is-active': !!debouncedQuery },
+      )
+    }
+    >
+      <div className="dropdown-trigger ">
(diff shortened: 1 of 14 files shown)


src/components/Dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,119 @@
+import React, { useCallback, useMemo, useState } from 'react';
+import './Dropdown.scss';
+import cn from 'classnames';
+import { debounce } from 'lodash';
```
YegorKorenkov, line 4: Please chek the linter error and try to fix
https://github.com/mate-academy/react_autocomplete/actions/runs/5495363361/jobs/10014713836?pr=86

src/components/Dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,119 @@
...
+  );
+
+  const handleResetQuery = useCallback(
+    () => {
+      setSelectedPerson(null);
+      setQuery('');
+      setDebouncedQuery('');
+    },
+    [],
+  );
```
YegorKorenkov, line 62: Why do you wrap each handler in useCallback? write why please

YegorKorenkov (changes requested):

src/components/Dropdown/Dropdown.tsx
adv555: fixed ✅ added lodash and fontawesome to dependenses

src/components/Dropdown/Dropdown.tsx
adv555: To prevent extra rendering in smaller components in Dropdown if something will be changed inside Dropdown, but forgot to split it.

src/components/Dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,80 @@
...
+import { DropdownMenu } from './DropdownMenu';
+
+type Props = {
+  people: Person[];
+  setSelectedPerson: (person: Person | null) => void;
+  delay: number;
+
+};
+
+export const Dropdown: React.FC<Props> = React.memo(
```
anstsot, line 16: do not use `memo` everywhere

anstsot (approved):
Good work, approving:sparkles:
But do not use `memo` everywhere, please. It can cause a lot of bugs

PR #92, opened 2023-07-11
autocomplete

Ir-ra:
[DEMO LINK](https://Ir-ra.github.io/react_autocomplete/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,14 +1,48 @@
-import React from 'react';
+import React, { useMemo, useState, useCallback } from 'react';
 import './App.scss';
+import debounce from 'lodash.debounce';
 import { peopleFromServer } from './data/people';
+import { Person } from './types/Person';
+import { People } from './components/People';
+
+const initialPeople: Person[] = peopleFromServer.map(person => ({
+  ...person,
+}));
 
 export const App: React.FC = () => {
-  const { name, born, died } = peopleFromServer[0];
+  const [query, setQuery] = useState('');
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -1,14 +1,48 @@
...
 export const App: React.FC = () => {
-  const { name, born, died } = peopleFromServer[0];
+  const [query, setQuery] = useState('');
+  const [appliedQuery, setAppliedQuery] = useState('');
+  const [selectedPerson, setSelectedPerson] = useState<Person | null>(null);
+
+  const applyQuery = useCallback(
+    debounce(setAppliedQuery, 1000),
+    [],
+  );
```
qoa11a, line 20: Don't use the `useCallback` without a reason.
You can read more about this hook here: https://react.dev/reference/react/useCallback

qoa11a (changes requested):

src/App.tsx
```diff
@@ -1,14 +1,45 @@
-import React from 'react';
+import React, { useMemo, useState } from 'react';
 import './App.scss';
+import debounce from 'lodash.debounce';
 import { peopleFromServer } from './data/people';
+import { Person } from './types/Person';
+import { People } from './components/People';
+
+const initialPeople: Person[] = peopleFromServer.map(person => ({
```
lerastarynets, line 8: The `.map` operation here is redundant, as you're spreading each person object into a new one without any modifications. Consider using the original `peopleFromServer` array instead.

PR #95, opened 2023-07-12
react_autocomplete

vintwp:
https://vintwp.github.io/react_autocomplete/


src/components/DropdownList/DropdownList.scss
```diff
@@ -0,0 +1,9 @@
+.dropdown {
+  &-item {
+    cursor: pointer;
+
+    &:hover p {
+      color: red !important;
```
volodymyr-soltys97, line 6: Avoid using '!important' in CSS. It increases the specificity of a rule and can make debugging more difficult. Try to use more specific selectors or re-arrange your CSS rules.

src/components/InputField/InputField.tsx
```diff
@@ -0,0 +1,45 @@
...
+export const InputField: React.FC<Props> = ({
+  delay,
+  onQuery,
+  selectedPerson,
+}) => {
+  const [query, setQuery] = useState('');
+
+  const applyQuery = useCallback(
+    debounce(onQuery, delay),
+    [],
```
volodymyr-soltys97, line 20: The dependencies array of useCallback is empty. This means the debounce function will not update if the props delay or onQuery change. You should include delay and onQuery in the array.

src/components/DropdownList/DropdownList.tsx
```diff
@@ -0,0 +1,33 @@
...
+    <div className="dropdown-menu" role="menu">
+      <div className="dropdown-content">
+        { peoples
+          ? peoples.map((man, index) => (
+            <div
+              key={`${man.name}-${man.born}`}
+              className="dropdown-item"
+              onClick={() => onSelect(man)}
+              onKeyDown={() => onSelect(man)}
+              role="button"
```
volodymyr-soltys97, line 23: It's not necessary to set role="button" on div if you are making it clickable. To improve accessibility, consider using a button instead of a div for clickable items.

src/components/DropdownList/DropdownList.tsx
```diff
@@ -0,0 +1,33 @@
+import { Person } from '../../types/Person';
+import './DropdownList.scss';
+
+type Props = {
+  peoples: Person[] | null;
```
volodymyr-soltys97, line 5: The prop 'peoples' might be better named as 'people', as 'people' is already plural.

src/components/DropdownList/DropdownList.tsx
```diff
@@ -0,0 +1,33 @@
...
+  return (
+    <div className="dropdown-menu" role="menu">
+      <div className="dropdown-content">
+        { peoples
+          ? peoples.map((man, index) => (
+            <div
+              key={`${man.name}-${man.born}`}
+              className="dropdown-item"
+              onClick={() => onSelect(man)}
+              onKeyDown={() => onSelect(man)}
```
volodymyr-soltys97, line 22: The onKeyDown event handler should not trigger onSelect function as it does not provide a specific key condition, it will fire on any key press. It's recommended to handle specific key events like 'Enter' or 'Space'.

src/App.tsx
```diff
@@ -1,56 +1,62 @@
...
-              <p className="has-text-link">Pieter Antone Haverbeke</p>
-            </div>
+      return name.toLowerCase().includes(query.toLowerCase());
+    });
 
-            <div className="dropdown-item">
-              <p className="has-text-danger">Elisabeth Haverbeke</p>
-            </div>
+    if (filtered.length === 0) {
+      return null;
```
volodymyr-soltys97, line 24: Returning `null` for `filteredPeoples` when there are no matches might be a problem. It would be better to return an empty array to prevent potential errors in the `DropdownList` component where you are likely iterating over `peoples`.

volodymyr-soltys97 (changes requested):
Good job
But let's make your code perfect

volodymyr-soltys97 (approved):
Nice work 🔥

PR #96, opened 2023-07-13
Develop

read-my-code:
[DEMO LINK](https://read-my-code.github.io/react_autocomplete/)

### DIFF
--- a/src/components/dropdown.tsx
+++ b/src/components/dropdown.tsx
@@ -1,0 +1,39 @@
+import React from 'react';
+import { Person } from '../types/Person';
+
+interface Props {
+  people: Person[];
+  onSelect?: (person: Person) => void;
+}
+
+export const Dropdown: React.FC<Props> = React.memo((
+  {
+    people,
+    onSelect = () => { },
+  },
+) => (
+  <div className="dropdown-menu" role="menu">
+    <div className="dropdown-content">
+      {!people.length ? (
+        <div className="dropdown-item">
+          <p>
+            No matching suggestions
+          </p>
+        </div>
+      ) : people.map(person => (
+        // eslint-disable-next-line jsx-a11y/click-events-have-key-events, jsx-a11y/no-static-element-interactions
+        <div
+          key={person.name}
+          className="dropdown-item"
+          onClick={() => onSelect(person)}
+        >
+          <p
+            className="has-text-link"
+          >
+            {person.name}
+          </p>
+        </div>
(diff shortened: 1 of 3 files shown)


src/components/dropdown.tsx
```diff
@@ -0,0 +1,39 @@
...
+  people: Person[];
+  onSelect?: (person: Person) => void;
+}
+
+export const Dropdown: React.FC<Props> = (
+  {
+    people,
+    onSelect = () => { },
+  },
+) => (
```
loralevitska, line 14:
```suggestion
export const Dropdown: React.FC<Props> = ({
    people,
    onSelect = () => { },
}) => (
```

src/components/dropdown.tsx
```diff
@@ -0,0 +1,39 @@
...
+  <div className="dropdown-menu" role="menu">
+    <div className="dropdown-content">
+      {!people.length ? (
+        <div className="dropdown-item">
+          <p>
+            No matching suggestions
+          </p>
+        </div>
+      ) : people.map(person => (
+        // eslint-disable-next-line jsx-a11y/click-events-have-key-events, jsx-a11y/no-static-element-interactions
```
loralevitska, line 24: Try to fix eslint rule and remove comment

loralevitska (changes requested):
Almost done) Let's improve

- make the search case-insensitive
image
image

src/components/dropdown.tsx
read-my-code: Там не можливо пофіксити, бо я задаю onClick  діву, а це не інтерактивний елемент

src/components/dropdown.tsx
```diff
@@ -0,0 +1,41 @@
...
+interface Props {
+  people: Person[];
+  onSelect?: (person: Person) => void;
+}
+
+export const Dropdown: React.FC<Props> = ({
+  people,
+  onSelect = () => { },
+}) => {
+  console.log('Rednder Dropdown');
```
loralevitska, line 13: Remove console

src/components/dropdown.tsx
loralevitska: I believe you) the idea was for you to rewrite it to a select with options or an input or some other interactive element. For this task, this is only a recommendation, but there are situations when it is necessary and it would be good if you knew how to do it

loralevitska (approved):
Well done)

PR #103, opened 2023-07-17
add task solution

VolodymyrPst:
https://volodymyrpst.github.io/react_autocomplete/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,14 +1,44 @@
-import React from 'react';
+import React, { useState, useMemo } from 'react';
 import './App.scss';
+import classNames from 'classnames';
 import { peopleFromServer } from './data/people';
+import { Person } from './types/Person';
 
 export const App: React.FC = () => {
-  const { name, born, died } = peopleFromServer[0];
+  const [text, setText] = useState('');
+  const [person, setPerson] = useState<Person | null>();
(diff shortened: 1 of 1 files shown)


anastasiiavorobiova (changes requested):
Great work! Added comments to fix
Screenshot 2023-07-17 at 16 40 21
Consider removing the empty dropdown content if there are no suggestions and the user wasn't interacting with dropdown

anastasiiavorobiova (changes requested):
Great work! Keep doing: 1) 
Screenshot 2023-07-17 at 18 19 47
If there are no matching suggestions it should be shown as a message in the dropdown 
2) Screenshot 2023-07-17 at 18 19 07
If the person is selected, the dropdown should be hidden

src/App.tsx
```diff
@@ -1,14 +1,47 @@
...
+          );
+        });
+
+      setSame(filtered);
+      setSearch(true);
+    }, 1000);
+  };
+
+  useEffect(() => {
+    if (text !== '') {
```
loralevitska, line 31:
```suggestion
    if (text) {
```

loralevitska (changes requested):
Almost perfect) Let's improve

`no matching suggestions` better write in dropdown or remove empty dropdown

image

PR #105, opened 2023-07-17
dropdown menu was implemented

2pasha:
[DEMO LINK](https://2pasha.github.io/react_autocomplete/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,10 +1,21 @@
 import React, { useCallback, useRef, useState } from 'react';
-import debounce from 'lodash.debounce';
 import classNames from 'classnames';
 
 import './App.scss';
 import { peopleFromServer } from './data/people';
 import { Person } from './types/Person';
+
+function debounce(callback: (...args: any) => void, delay: number) {
+  let timerId = 0;
+
+  return (...args: any) => {
+    window.clearTimeout(timerId);
+
+    timerId = window.setTimeout(() => {
+      callback(...args);
+    }, delay);
+  };
+}
 
 export const App: React.FC = () => {
   const [query, setQuery] = useState('');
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -1,54 +1,92 @@
-import React from 'react';
+import React, { useCallback, useRef, useState } from 'react';
+import debounce from 'lodash.debounce';
```
mykhalenych, line 2: better to create your own debounce

src/App.tsx
```diff
@@ -1,54 +1,92 @@
...
+    prevQuery.current = event.target.value;
+  };
+
+  const selectPerson = (person: Person) => {
+    setSelectedPerson(person);
+    setQuery(person.name);
+  };
+
+  const handleOnBlurInput = () => {
+    setTimeout(() => {
```
mykhalenych, line 38: why do you need timeout when you already use debounce?

src/App.tsx
```diff
@@ -1,54 +1,92 @@
...
+  const [selectedPerson, setSelectedPerson] = useState<Person | null>(null);
+  const [openMenu, setOpenMenu] = useState(false);
+
+  const apllyQuery = useCallback(
+    debounce(setAppliedQuery, 1000),
+    [],
+  );
+
+  const handleQueryCahnge = (event: React.ChangeEvent<HTMLInputElement>) => {
+    setQuery(event.target.value);
```
mykhalenych, line 22: you did setQuery after apllyQuery, and after check if input value the same as previous search, looks strange, better to check it before

mykhalenych (changes requested):

etojeDenys (approved):
lgtm

PR #106, opened 2023-07-17
finished task

artkas03:
[DEMO LINK](https://artkas03.github.io/react_autocomplete/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,14 +1,73 @@
-import React from 'react';
+import React, { useCallback, useMemo, useState } from 'react';
 import './App.scss';
+import debounce from 'lodash.debounce';
 import { peopleFromServer } from './data/people';
+import { Person } from './types/Person';
+import { PeopleList } from './components/PeopleList';
+
+type FilterParams = {
+  query: string;
+};
+
+function filterPeople(people: Person[], { query }: FilterParams): Person[] {
+  let resultPeople = [...people];
+  const normalazidQuery = query.toLowerCase();
+
+  if (query) {
+    resultPeople = resultPeople.filter((person) => {
+      const normalazidName = person.name.toLowerCase();
+
+      return normalazidName.includes(normalazidQuery);
+    });
+  }
+
+  return resultPeople;
+}
 
 export const App: React.FC = () => {
-  const { name, born, died } = peopleFromServer[0];
+  const [query, setQuery] = useState('');
+  const [preparedQuery, setPreparedQueary] = useState('');
+  const setQueryToFilter = useCallback(
+    debounce(setPreparedQueary, 1000),
+    [],
+  );
+
+  const [selectedPerson, setSelectedPerson] = useState<Person | null>(null);
+
+  const visiblePeople = useMemo(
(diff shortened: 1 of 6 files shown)


src/App.tsx
```diff
@@ -1,14 +1,73 @@
...
+      const normalazidName = person.name.toLowerCase();
+
+      return normalazidName.includes(normalazidQuery);
+    });
+  }
+
+  return resultPeople;
+}
 
 export const App: React.FC = () => {
```
GGsmou, line 27: The task was to create a separate component `Autocomplete` and move most functionality there

src/App.tsx
```diff
@@ -1,14 +1,73 @@
...
-  const { name, born, died } = peopleFromServer[0];
+  const [query, setQuery] = useState('');
+  const [preparedQuery, setPreparedQueary] = useState('');
+  const setQueryToFilter = useCallback(
+    debounce(setPreparedQueary, 1000),
```
GGsmou, line 31: Debounce must be customizable, so it better to use variable for these

src/components/PeopleList/PeopleList.tsx
```diff
@@ -0,0 +1,31 @@
+/* eslint-disable jsx-a11y/no-static-element-interactions */
+/* eslint-disable jsx-a11y/click-events-have-key-events */
+import React from 'react';
+import cn from 'classnames';
+import { Person } from '../../types/Person';
+
+type Props = {
+  people: Person[],
+  onClick: (person: Person) => void,
```
GGsmou, line 9: I think it's better to name `onClick` as `onSelected`

GGsmou (approved):
GJ, approving!
The task was to create a separate component for full functionality, not only for rendering People List, but I think that's not crucial)

src/components/PeopleList/PeopleList.tsx
And678: Yes

src/App.tsx
And678: Yup, Implement the Autocomplete component

src/App.tsx
And678: + here as well

src/App.tsx
```diff
@@ -1,14 +1,73 @@
...
+  function handleOnClick(person: Person): void {
+    setSelectedPerson(person);
+
+    setQuery(person.name.toString());
+  }
+
+  const dropdownList = (
+    <PeopleList
+      people={visiblePeople}
+      onClick={(newPerson) => handleOnClick(newPerson)}
```
And678, line 57:
```suggestion
      onClick={handleOnClick}
```

src/App.tsx
```diff
@@ -1,14 +1,73 @@
...
+    setQuery(newQuery);
+    setQueryToFilter(newQuery);
+    setSelectedPerson(null);
+  }
+
+  function handleOnClick(person: Person): void {
+    setSelectedPerson(person);
+
+    setQuery(person.name.toString());
+  }
```
And678, line 52: Wrap this in useCallback, because this is passed to memo'ed component

And678 (changes requested):

PR #107, opened 2023-07-18
react_autocomplete

hrynkevych:
[DEMO LINK](https://hrynkevych.github.io/react_autocomplete/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,55 +1,68 @@
-import React from 'react';
-import './App.scss';
+import React, { useCallback, useMemo, useState } from 'react';
+import debounce from 'lodash/debounce';
+import cn from 'classnames';
+
+import { DropdownMenu } from './Components/DropdownMenu';
 import { peopleFromServer } from './data/people';
+import { Person } from './types/Person';
+
+import './App.scss';
 
 export const App: React.FC = () => {
-  const { name, born, died } = peopleFromServer[0];
+  const [query, setQuery] = useState('');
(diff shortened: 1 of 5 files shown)


src/App.tsx
```diff
@@ -1,55 +1,68 @@
-import React from 'react';
-import './App.scss';
+import React, { useCallback, useMemo, useState } from 'react';
+import debounce from 'lodash/debounce';
```
mykhalenych, line 2: try to create your own debounce function

src/App.tsx
```diff
@@ -1,55 +1,68 @@
...
+  }, [appliedQuery]);
+
+  const applyQuery = useCallback(
+    debounce(setAppliedQuery, 1000),
+    [],
+  );
+
+  const queryChangeHandler = (event: React.ChangeEvent<HTMLInputElement>) => {
+    setQuery(event.target.value);
+    applyQuery(event.target.value);
```
mykhalenych, line 29: what do you do here?

mykhalenych (changes requested):

src/App.tsx
hrynkevych: ok, got it, i`ll fix this

VitaliyBondarenko1982 (changes requested):
this point from task description not impemented
- show the list of all people when input is focused but empty ([Check this](https://mui.com/material-ui/react-autocomplete/#combo-box) to see how it can work)

PR #109, opened 2023-07-18
add task solution

irynabolbat:
[DEMO LINK](https://irynabolbat.github.io/react_autocomplete/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,56 +1,95 @@
-import React from 'react';
+import React, { useCallback, useMemo, useState } from 'react';
+import debounce from 'lodash.debounce';
 import './App.scss';
+import classNames from 'classnames';
+
 import { peopleFromServer } from './data/people';
+import { Person } from './types/Person';
+import { PeopleList } from './components/PeopleList/PeopleList';
 
 export const App: React.FC = () => {
-  const { name, born, died } = peopleFromServer[0];
+  const delay = 1000;
+  const allPeople: Person[] = peopleFromServer;
+
+  const [selectedPerson, setSelectedPerson] = useState<Person | null>(null);
+  const [query, setQuery] = useState<string>('');
+  const [showList, setShowList] = useState<boolean>(false);
+
+  const filteredPeople = useMemo(() => {
+    return allPeople.filter(
+      person => person.name.toLowerCase().includes(query.toLowerCase().trim()),
+    );
+  }, [query, allPeople]);
+
+  const applySearch = useCallback(
+    debounce(value => {
+      setQuery(value);
+    }, delay),
+    [delay],
+  );
+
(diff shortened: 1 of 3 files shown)


src/App.tsx
```diff
@@ -1,56 +1,95 @@
...
+    debounce(value => {
+      setQuery(value);
+    }, delay),
+    [delay],
+  );
+
+  const handleInputChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+    setQuery(event.target.value);
+    setShowList(true);
+    applySearch(event.target.value);
```
dianadiv, line 34: When you instantly change the query, you have an array of people filtered without delay. I think you can create another query that will be set with a delay in debounce and filter the array. Or just delete 32 line

dianadiv (changes requested):
Good job, but the list of people is displayed without delay

dianadiv (approved):

src/App.tsx
```diff
@@ -1,56 +1,92 @@
...
+  const handleInputChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+    setQuery(event.target.value);
+    applySearch(event.target.value);
+  };
+
+  const handleSelectPerson = (person: Person) => {
+    setQuery(person.name);
+    setApplyQuery(person.name);
+    setSelectedPerson(person);
+  };
```
And678, line 38: Wrap this in useCallback, so memo'ed `PeopleList` can avoid extra rerender when this function is passed

And678 (changes requested):
List is not closed after selecting:
image

src/App.tsx
```diff
@@ -1,56 +1,92 @@
...
 import './App.scss';
+import classNames from 'classnames';
+
 import { peopleFromServer } from './data/people';
+import { Person } from './types/Person';
+import { PeopleList } from './components/PeopleList/PeopleList';
 
 export const App: React.FC = () => {
-  const { name, born, died } = peopleFromServer[0];
+  const delay = 1000;
```
anstsot, line 11: better to create out of component

anstsot (changes requested):
Good job:four_leaf_clover:
Try to fix few comments and create a new component to meet one condition of the task:
- the delay should be customizable via props;

src/App.tsx
```diff
@@ -1,56 +1,92 @@
...
+    [delay],
+  );
+
+  const handleInputChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+    setQuery(event.target.value);
+    applySearch(event.target.value);
+  };
+
+  const handleSelectPerson = useCallback((person: Person) => {
+    setQuery('');
```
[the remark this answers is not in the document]
anstsot: still not fixed

anstsot (changes requested):
Try to fix previous comments. Create Autocomplete component and propagate `delay` as prop

PR #119, opened 2023-07-19
add task solution

AlyonaSarapina:
[DEMO LINK](https://alyonasarapina.github.io/react_autocomplete/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,56 +1,78 @@
-import React from 'react';
+import classNames from 'classnames';
+import React, {
+  useMemo,
+  useState,
+  useCallback,
+} from 'react';
 import './App.scss';
+import { PeopleList } from './components/PeopleLIst';
 import { peopleFromServer } from './data/people';
+import { Person } from './types/Person';
 
-export const App: React.FC = () => {
-  const { name, born, died } = peopleFromServer[0];
+interface Props {
+  delay: number,
+}
+
+function debounce(callback: (...args: string[]) => void, delay: number) {
+  let timerId = 0;
+
+  return (...args: string[]) => {
+    window.clearTimeout(timerId);
+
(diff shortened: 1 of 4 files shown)


src/App.tsx
```diff
@@ -1,56 +1,78 @@
...
+  let timerId = 0;
+
+  return (...args: string[]) => {
+    window.clearTimeout(timerId);
+
+    timerId = window.setTimeout(() => {
+      callback(...args);
+    }, delay);
+  };
+}
```
outofs, line 26: Personally, I think it will be better to use `lodash.debounce` library

src/App.tsx
```diff
@@ -1,56 +1,78 @@
...
-              <p className="has-text-danger">Elisabeth Hercke</p>
-            </div>
-          </div>
-        </div>
+        {appliedQuery && (
+          <PeopleList
+            people={filteredPeople}
+            onSelect={handleSelect}
+          />
+        )}
```
outofs, line 75: You're already using functionality with classes (lines 56-60) for activating drop down list, so there is no need of using conditional rendering

outofs (approved):
Well done! Please, take my some advises below

src/components/PeopleLIst.tsx
```diff
@@ -0,0 +1,33 @@
+import React from 'react';
+import { Person } from '../types/Person';
+
+interface Props {
+  people: Person[];
+  onSelect?: (person: Person) => void;
+}
+
+export const PeopleList: React.FC<Props> = ({
```
And678, line 9: Use `React.memo` here to prevent extra renders

And678 (changes requested):
Requirement not implemented: show the list of all people when input is focused but empty

anstsot (changes requested):
Let's add all requirements:

PR #120, opened 2023-07-19
code

liza-strykharchuk:
[DEMO LINK](https://liza-strykharchuk.github.io/react_autocomplete/)

### DIFF
--- a/src/conponents/Dropdown.tsx
+++ b/src/conponents/Dropdown.tsx
@@ -1,0 +1,35 @@
+import React from 'react';
+import classNames from 'classnames';
+import { Person } from '../types/Person';
+
+type Props = {
+  people: Person[];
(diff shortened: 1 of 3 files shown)


src/conponents/Dropdown.tsx
```diff
@@ -0,0 +1,35 @@
...
+};
+
+export const Dropdown: React.FC<Props> = React.memo(({
+  people,
+  onSelected,
+}) => {
+  return (
+    <div className="dropdown-content">
+      {people.map(person => (
+        <div className="dropdown-item" key={person.name}>
```
vlad-demchuk, line 17: do not use name as keys

vlad-demchuk (changes requested):
GJ!

volodymyr-soltys97 (approved):

PR #121, opened 2023-07-19
solution

anya-shults:
- [DEMO LINK](https://anya-shults.github.io/react_autocomplete/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,57 +1,34 @@
-import React from 'react';
+import React, { useMemo, useState } from 'react';
+
 import './App.scss';
+import { Dropdown } from './components';
 import { peopleFromServer } from './data/people';
+import { Person } from './types/Person';
 
 export const App: React.FC = () => {
-  const { name, born, died } = peopleFromServer[0];
+  const [onSelected, setOnSelected] = useState<Person | null>(null);
+
+  const selectedPerson = useMemo(() => {
+    if (onSelected) {
+      const { name, born, died } = onSelected;
+
+      return `${name} (${born} - ${died})`;
+    }
+
+    return 'No selected person';
+  }, [onSelected]);
 
   return (
     <main className="section">
       <h1 className="title">
-        {`${name} (${born} = ${died})`}
+        {selectedPerson}
       </h1>
 
-      <div className="dropdown is-active">
-        <div className="dropdown-trigger">
-          <input
-            type="text"
-            placeholder="Enter a part of the name"
-            className="input"
-          />
-        </div>
-
-        <div className="dropdown-menu" role="menu">
-          <div className="dropdown-content">
-            <div className="dropdown-item">
-              <p className="has-text-link">Pieter Haverbeke</p>
-            </div>
-
-            <div className="dropdown-item">
-              <p className="has-text-link">Pieter Bernard Haverbeke</p>
-            </div>
-
-            <div className="dropdown-item">
-              <p className="has-text-link">Pieter Antone Haverbeke</p>
-            </div>
-
-            <div className="dropdown-item">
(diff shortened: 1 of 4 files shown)


src/App.tsx
```diff
@@ -1,57 +1,34 @@
...
+import React, { useMemo, useState } from 'react';
+
 import './App.scss';
+import { Dropdown } from './components';
 import { peopleFromServer } from './data/people';
+import { Person } from './types/Person';
 
 export const App: React.FC = () => {
-  const { name, born, died } = peopleFromServer[0];
+  const [onSelected, setOnSelected] = useState<Person | null>(null);
```
olena-ole, line 9: it's better to call these variables selectedPerson and setSelectedPerson

src/App.tsx
```diff
@@ -1,57 +1,34 @@
...
+
+  const selectedPerson = useMemo(() => {
+    if (onSelected) {
+      const { name, born, died } = onSelected;
+
+      return `${name} (${born} - ${died})`;
+    }
+
+    return 'No selected person';
+  }, [onSelected]);
```
olena-ole, line 19: You don't have to memoize this value cause you don't pass it as props, instead, you can create a variable for the title text, e.g.:
```js
const titleText = selectedPerson
    ? `${selectedPerson.name} (${selectedPerson.born} = ${selectedPerson.died})`
    : 'No selected person';

File: src/App.tsx [hunk 3]
```diff
@@ -1,57 +1,34 @@
...
 
   return (
     <main className="section">
       <h1 className="title">
-        {`${name} (${born} = ${died})`}
+        {selectedPerson}
```
olena-ole, line 24: and then here you can use {titleText}

src/components/Dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,91 @@
...
+  const handleClick = useCallback((
+    person: Person,
+    event: React.MouseEvent<HTMLAnchorElement, MouseEvent>,
+  ) => {
+    event.preventDefault();
+
+    setValue(person.name);
+    setOnSelected(person);
+    setAppliedValue('');
+  }, []);
```
olena-ole, line 50: I would recommend moving this logic to App component (App component receives delay as props, while dropdown receives only memoized filteredPeople array and onSelect function), then your memoization will make sense. Here you rerender your component on every input change

src/components/Dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,91 @@
...
+
+    setValue(normalisedEvent);
+    applyValue(normalisedEvent);
+  };
+
+  const handleClick = useCallback((
+    person: Person,
+    event: React.MouseEvent<HTMLAnchorElement, MouseEvent>,
+  ) => {
+    event.preventDefault();
```
olena-ole, line 45: You don't need to call event.preventDefault() here, it's just a click

olena-ole (changes requested):
Good start! Let's add some improvements to make your code optimized!

src/components/Dropdown/Dropdown.tsx
olena-ole: My bad, it's a link)

olena-ole (approved):
Great Job, approve!🎂🎉🎈

anstsot (changes requested):
Let's add all requirements:
- make sure that filter won't be called if user entered spaces only;

https://github.com/mate-academy/react_autocomplete/assets/22824431/f070a88f-d6ca-4f8a-a4b1-d4ea8a6c549d

anstsot (approved):
Good work, approving:sparkles:

PR #123, opened 2023-07-19
solution

ihorhladkov:
[DEMO LINK](https://ihorhladkov.github.io/react_autocomplete/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,57 +1,64 @@
-import React from 'react';
+import debounce from 'lodash.debounce';
+import React, { useMemo, useState, useCallback } from 'react';
 import './App.scss';
+import classNames from 'classnames';
 import { peopleFromServer } from './data/people';
+import { Person } from './types/Person';
+import { Dropdown } from './components/Dropdown';
+
+const preparedPeople = (
+  people: Person[],
+  query: string,
+) => {
+  const peopleCopy = [...people].filter(person => (
+    person.name.toLowerCase().includes(query.toLowerCase())
+  ));
+
+  return peopleCopy;
+};
 
 export const App: React.FC = () => {
-  const { name, born, died } = peopleFromServer[0];
+  const [query, setQuery] = useState<string>('');
+  const [appliedQuery, setAppliedQuery] = useState<string>('');
+  const people = useMemo(() => (
+    preparedPeople(peopleFromServer, appliedQuery)),
+  [appliedQuery]);
+  const [selectedUser, setSelectedUser] = useState<Person | null>(null);
+
+  const applyQuery = useCallback(
+    debounce(setAppliedQuery, 1000),
+    [],
+  );
+
+  const handleQueryChange = (name: string) => {
+    setQuery(name);
+    applyQuery(name);
+  };
 
   return (
     <main className="section">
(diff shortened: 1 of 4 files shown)


src/App.tsx
```diff
@@ -1,57 +1,64 @@
...
+      <Dropdown
+        people={people}
+        query={query}
+        selectedUser={selectedUser}
+        setQuery={setQuery}
+        setSelectedUser={setSelectedUser}
+        handleQueryChange={handleQueryChange}
+        appliedQuery={appliedQuery}
+        setAppliedQuery={setAppliedQuery}
+      />
```
dianadiv, line 61: I think it would be better to leave the input in the App component to avoid passing so many props.

src/components/Dropdown.tsx
```diff
@@ -0,0 +1,75 @@
...
+        <input
+          type="text"
+          placeholder="Enter a part of the name"
+          className="input"
+          value={query}
+          onChange={(event) => handleQueryChange(event.target.value)}
+        />
+      </div>
+
+      {appliedQuery && (
```
dianadiv, line 47: according to readme : hide suggestions on text change. You can add a check if appliedQuery === query to avoid displaying a list of people while typing

dianadiv (approved):
Good job, approved

src/App.tsx
```diff
@@ -1,57 +1,64 @@
...
 
-            <div className="dropdown-item">
-              <p className="has-text-link">Pieter Antone Haverbeke</p>
-            </div>
+export const App: React.FC = () => {
+  const [query, setQuery] = useState<string>('');
+  const [appliedQuery, setAppliedQuery] = useState<string>('');
+  const people = useMemo(() => (
+    preparedPeople(peopleFromServer, appliedQuery)),
+  [appliedQuery]);
```
anstsot, line 25:
```suggestion
  const people = useMemo(() => (
    preparedPeople(peopleFromServer, appliedQuery)),
    [appliedQuery],
  );
```

anstsot (changes requested):
Good work:four_leaf_clover:
Try to add one more condition of the task:
- hide suggestions on text change;

https://github.com/mate-academy/react_autocomplete/assets/22824431/934d1e4b-de13-465d-8612-1741a0080b01

src/App.tsx
ihorhladkov: Linter does not allow this(

src/App.tsx
```diff
@@ -1,57 +1,64 @@
...
+  const [selectedUser, setSelectedUser] = useState<Person | null>(null);
 
-            <div className="dropdown-item">
-              <p className="has-text-danger">Elisabeth Haverbeke</p>
-            </div>
+  const applyQuery = useCallback(
+    debounce(setAppliedQuery, 1000),
```
And678, line 29: No reason to have separate applied query, just call preparePeople here

And678 (changes requested):
Fix your tests by pushing the updated `package.json` (you installed lodash.debounce). 

Also, this requirement is not done `show the list of all people when input is focused but empty ([Check this](https://mui.com/material-ui/react-autocomplete/#combo-box) to see how it can work)`

PR #128, opened 2023-07-20
add task solution

AndriiYelieva:
[DEMO ----](https://AndriiYelieva.github.io/react_autocomplete/)


src/App.tsx
```diff
@@ -1,56 +1,66 @@
-import React from 'react';
+import React, { useCallback, useMemo, useState } from 'react';
+import debounce from 'lodash.debounce';
```
IvanFesenko, line 2: Try to write own debounce

src/App.tsx
```diff
@@ -1,56 +1,66 @@
...
+  const handleQueryChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+    applyQuery(event.target.value);
+  };
+
+  const filteredPeople = useMemo(() => {
+    return peopleFromServer.filter(person => (
+      person.name.toLowerCase()
+        .includes(appliedQuery.trim().toLocaleLowerCase())
+    ));
+  }, [appliedQuery]);
```
IvanFesenko, line 33: No need to filter in case query is empty

IvanFesenko (changes requested):
Good job. Check my comments

src/App.tsx
```diff
@@ -1,56 +1,87 @@
...
+  const filteredPeople = useMemo(() => {
+    if (appliedQuery.trim()) {
+      return peopleFromServer.filter(person => (
+        person.name.toLowerCase()
+          .includes(appliedQuery.trim().toLocaleLowerCase())
+      ));
+    }
+
+    return peopleFromServer;
+  }, [appliedQuery]);
```
IvanFesenko, line 53:
```suggestion
  const filteredPeople = useMemo(() => {
    const trimmedValue = appliedQuery.trim();
  
    if (trimmedValue) {
      return peopleFromServer.filter(person => (
        person.name.toLowerCase()
          .includes(trimmedValue.toLocaleLowerCase())
      ));
    }

    return peopleFromServer;
  }, [appliedQuery]);
```

IvanFesenko (approved):

PR #130, opened 2023-07-20
solution

artem-ivashchenko:
[DEMO LINK](https://artem-ivashchenko.github.io/react_autocomplete/)

### DIFF
--- a/src/components/Dropdown.tsx
+++ b/src/components/Dropdown.tsx
@@ -1,0 +1,97 @@
+import classNames from 'classnames';
+import React, { useState, useMemo, useCallback } from 'react';
+
+import { Person } from '../types/Person';
+import { peopleFromServer } from '../data/people';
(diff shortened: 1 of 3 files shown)


src/components/Dropdown.tsx
```diff
@@ -0,0 +1,97 @@
...
+  const handleButtonClick = (
+    event: React.MouseEvent<HTMLAnchorElement, MouseEvent>,
+    person: Person,
+  ) => {
+    event.preventDefault();
+    onSelected(person);
+    setQuery(person.name);
+  };
+
+  const hasDropdown = React.useMemo(() => {
```
witflash, line 44: No need to use `React.useMemo`. You can use directly `useMemo` as you did for other places.

witflash (approved):
Well done 👍

PR #132, opened 2023-07-20
feat: dropdown provided

OKhalanskyi:
[DEMO LINK](https://OKhalanskyi.github.io/react_autocomplete/)


anstsot (changes requested):
Let's fix filtering (works incorrect) and add all requirements:
- the `delay` should be customizable via props (create separate component);
- make sure that filter won't be called if user entered spaces only;

https://github.com/mate-academy/react_autocomplete/assets/22824431/bae1bff2-38cc-4332-84d5-735a05831edd

PR #137, opened 2023-07-22
solution

IzzetYusufov:
[DEMO LINK](https://IzzetYusufov.github.io/react_autocomplete/)


TarasHoliuk:
Don't show an empty list by default - you should show it only if smth is typed with `No matching suggestions` message if it's true, or list of suggestions
image

src/App.tsx
```diff
@@ -1,54 +1,74 @@
...
+  );
 
   return (
     <main className="section">
       <h1 className="title">
-        {`${name} (${born} = ${died})`}
+        {selectedPerson === null ? (
+          'No selected person'
+        ) : (
+          `${selectedPerson.name} (${selectedPerson.born} = ${selectedPerson.died})`
```
TarasHoliuk, line 43:
```suggestion
          `${selectedPerson.name} (${selectedPerson.born} - ${selectedPerson.died})`
```

src/App.tsx
```diff
@@ -1,54 +1,74 @@
...
 import './App.scss';
 import { peopleFromServer } from './data/people';
+import { PersonList } from './components/PersonList';
+import { Person } from './types/Person';
 
 export const App: React.FC = () => {
-  const { name, born, died } = peopleFromServer[0];
+  const [selectedPerson, setSelectedPerson] = useState<Person | null>(null);
+  const [query, setQuery] = useState<string>('');
+  const [debounceQuery, setDebounceQuery] = useState<string>('');
```
TarasHoliuk, line 12:
```suggestion
  const [query, setQuery] = useState('');
  const [debounceQuery, setDebounceQuery] = useState('');
```
it's defined by default since you pass string as initial value

src/App.tsx
```diff
@@ -1,54 +1,74 @@
...
-              <p className="has-text-danger">Petronella de Decker</p>
-            </div>
-
-            <div className="dropdown-item">
-              <p className="has-text-danger">Elisabeth Hercke</p>
-            </div>
+            {filteringPersons.length > 0 ? (
+              debounceQuery && (
+                <PersonList
+                  persons={filteringPersons}
```
TarasHoliuk, line 63: try to do this:
image

ask in the chat if you need help

TarasHoliuk (changes requested):

volodymyr-soltys97 (approved):
Well done 👍

PR #140, opened 2023-07-25
solution

ShamievDavid:
[DEMO LINK](https://ShamievDavid.github.io/react_autocomplete/)

### DIFF
--- a/src/components/Autocompolete/Autocomplete.tsx
+++ b/src/components/Autocompolete/Autocomplete.tsx
@@ -1,0 +1,75 @@
+import React, { useState, useCallback } from 'react';
+import debounce from 'lodash.debounce';
+import { Person } from '../../types/Person';
+import { DropdownItem } from '../DropdownItem';
+
+type Props = {
+  people: Person [];
+  setPerson: (person: Person | null) => void;
+  setAppliedQuery: (e: string) => void;
+  delay: number;
+  selectedPerson: Person | null;
+};
+
+export const Autocomplete: React.FC<Props> = ({
+  people,
+  setPerson,
+  delay,
+  setAppliedQuery,
+  selectedPerson,
+}) => {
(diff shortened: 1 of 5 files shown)


src/components/DropdownItem/DropdownItem.tsx
```diff
@@ -0,0 +1,31 @@
+import React from 'react';
+import { Person } from '../../types/Person';
+
+type Props = {
+  person: Person | null;
```
VitaliyBondarenko1982, line 5: why person can be null. it prop comes from mapping people and looks like always will be Person

src/components/DropdownItem/DropdownItem.tsx
```diff
@@ -0,0 +1,31 @@
...
+  };
+
+  return (
+    // eslint-disable-next-line react/jsx-filename-extension, jsx-a11y/click-events-have-key-events, jsx-a11y/no-noninteractive-element-interactions, jsx-a11y/anchor-is-valid
+    <a
+      href="#"
+      className="dropdown-item has-text-link"
+      onClick={onPersonChange}
+    >
+      {person && person.name}
```
VitaliyBondarenko1982, line 28:
```suggestion
      {person.name}
```
check type of person prop. looks like person never be null

src/components/Autocompolete/Autocomplete.tsx
```diff
@@ -0,0 +1,75 @@
+import React, { useState, useCallback } from 'react';
+import debounce from 'lodash.debounce';
```
VitaliyBondarenko1982, line 2: try to write your own debounce function

VitaliyBondarenko1982 (changes requested):
Good job.
Needs some improvements

IvanFesenko (approved):

PR #144, opened 2023-07-27
Develop

dudar-y:
[DEMO LINK](https://dudar-y.github.io/react_autocomplete/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,56 +1,85 @@
-import React from 'react';
+import React, {
+  useCallback, useMemo, useState,
+} from 'react';
+
+import cn from 'classnames';
 import './App.scss';
+import { DropdownForm } from './components/DropdownForm/DropdownForm';
+import { DropdownMenu } from './components/DropdownMenu/DropdownMenu';
 import { peopleFromServer } from './data/people';
+import { Person } from './types/Person';
 
 export const App: React.FC = () => {
-  const { name, born, died } = peopleFromServer[0];
+  const [querry, setQuerry] = useState('');
+  const [appliedQuerry, setAppliedQuerry] = useState('');
+  const [isDropdownFocus, setIsDropdownFocus] = useState(false);
+  const [selectedPerson, setSelectedPerson] = useState<Person | null>(null);
+
+  const choosePerson = useCallback((newPerson: Person) => {
+    setSelectedPerson(newPerson);
+    setIsDropdownFocus(false);
+    setAppliedQuerry(newPerson.name.toString());
+  }, []);
+
+  const handleReset = useCallback(() => {
+    setSelectedPerson(null);
+    setQuerry('');
+    setAppliedQuerry('');
+  }, []);
+
+  const visablePeople = useMemo(() => {
+    return peopleFromServer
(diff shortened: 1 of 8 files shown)


etojeDenys (changes requested):
1. by default must be closed
image

src/App.tsx
```diff
@@ -1,56 +1,85 @@
...
+        {title}
+        <button
+          onClick={handleReset}
+          type="button"
+          className={cn(
+            'button_delete delete is-medium',
+            { 'is-hidden': !selectedPerson },
+          )}
+        >
+          {' '}
```
VitaliyBodnar, line 52: why do we need empty text ?

src/App.tsx
```diff
@@ -1,56 +1,85 @@
...
+
+        <button
+          onClick={() => setIsDropdownFocus(false)}
+          type="button"
+          className={cn(
+            'button_hidden delete is-medium',
+            { 'is-hidden': selectedPerson || !isDropdownFocus },
+          )}
+        >
+          {' '}
```
VitaliyBodnar, line 75: why do we need empty text ?

src/components/DropdownForm/DropdownForm.tsx
```diff
@@ -0,0 +1,56 @@
...
+
+export const DropdownForm: React.FC<Props> = ({
+  querry,
+  onQuerry,
+  onAppliedQuerry = () => { },
+  selectedPerson,
+  onDropdownFocus = () => { },
+}) => {
+  const [defoultImputFocus, setDefoultImputFocus] = useState(false);
+  const ApplyQuerry = useCallback(debounce(onAppliedQuerry, 1000), []);
```
VitaliyBodnar, line 23: the constant name should be in camelCase or Upper Upper
```suggestion
  const applyQuerry = useCallback(debounce(onAppliedQuerry, 1000), []);
```

src/components/DropdownForm/DropdownForm.tsx
```diff
@@ -0,0 +1,56 @@
...
+  useEffect(() => {
+    if (input.current && !selectedPerson && defoultImputFocus) {
+      input.current.focus();
+    }
+  }, [selectedPerson]);
+
+  const handleImputFocus = useCallback(() => {
+    onDropdownFocus(true);
+    setDefoultImputFocus(true);
+  }, []);
```
VitaliyBodnar, line 36: don't see any sense to use useCallback here, it can be just a function

VitaliyBodnar (changes requested):
Almost done!

src/App.tsx
dudar-y: I didn’t read about aria-label :(

src/App.tsx
```diff
@@ -1,56 +1,83 @@
-import React from 'react';
+import React, {
+  useCallback, useMemo, useState,
+} from 'react';
```
loralevitska, line 3:
```suggestion
import React, {
  useCallback, 
  useMemo, 
  useState,
} from 'react';
```

loralevitska (approved):
Good job!

PR #149, opened 2023-07-29
Solution

artyomwhite:
[DEMO LINK](https://artyomwhite.github.io/react_autocomplete/)


VitaliyBondarenko1982 (changes requested):
Hide suggestions
chrome-capture-2023-6-29 (2)
 when input lose focus (`onBlur` event)

etojeDenys (changes requested):
i cannot select user

PR #161, opened 2023-08-07
task solution

Bonum-taurus:
https://Bonum-taurus.github.io/react_autocomplete/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -5,11 +5,11 @@ import { peopleFromServer } from './data/people';
 import { Person } from './types/Person';
 
 export const App: React.FC = () => {
-  const [hasClick, setHasClick] = useState(false);
+  const [hasClick, setHasClick] = useState<boolean>(false);
   const [person, setPerson] = useState<Person | null>(null);
-  const [focus, setFocus] = useState(false);
-  const [queryOrPerson, setQueryOrPerson] = useState('');
-  const [aplliedQuery, setAplliedQuery] = useState('');
+  const [focus, setFocus] = useState<boolean>(false);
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -22,33 +64,20 @@ export const App: React.FC = () => {
...
-
-            <div className="dropdown-item">
-              <p className="has-text-danger">Petronella de Decker</p>
-            </div>
-
-            <div className="dropdown-item">
-              <p className="has-text-danger">Elisabeth Hercke</p>
-            </div>
+            {copyPeople.map(p => (
+              <a
```
volodymyr-soltys97, line 68: Use a destructuring here, and use more descriptive name
```suggestion
            {copyPeople.map(person => (
              <a
```

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code

anastasiiavorobiova (changes requested):
Fantastic work! Consider printing 'no matching suggestions here if there are no matches instead of keeping it empty: 
Screenshot 2023-08-08 at 16 44 49

src/App.tsx
```diff
@@ -22,33 +64,20 @@ export const App: React.FC = () => {
...
-
-            <div className="dropdown-item">
-              <p className="has-text-danger">Petronella de Decker</p>
-            </div>
-
-            <div className="dropdown-item">
-              <p className="has-text-danger">Elisabeth Hercke</p>
-            </div>
+            {copyPeople.map(man => (
+              <a
```
volodymyr-soltys97, line 68: You can use a destructuring for `man`

volodymyr-soltys97 (approved):
Well done 🔥

PR #163, opened 2023-08-11
add task solution

vovaAlieksieiev:
[DEMO LINK](https://vovaAlieksieiev.github.io/react_autocomplete/)


volodymyr-soltys97 (changes requested):
Demo link, you need to make a deploy
image

volodymyr-soltys97 (changes requested):
Need to show selected use in the field
image

PR #164, opened 2023-08-11
autocomplete

Dyakovich-R:
[DEMO LINK](https://Dyakovich-R.github.io/react_autocomplete/)


volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your task
1. By default, the text that there are no matches should not be shown, it should be shown only when the user enters data in the field
image

2. The selected user should be shown in the input field

image

maxim2310 (changes requested):
Good job, but find some bugs for you, fix it plz: 
1. image
2.when i selected should not be displayed "No matching suggestions"
image
3. when i selected user i cant remove it or edit or clear field

PR #165, opened 2023-08-13
add solution

denis-kononenko:
[DEMO LINK](https://denis-kononenko.github.io/react_autocomplete/)

### DIFF
--- a/src/components/Autocomplete/Autocomplete.tsx
+++ b/src/components/Autocomplete/Autocomplete.tsx
@@ -1,0 +1,101 @@
+import {
+  useCallback,
+  useEffect,
+  useMemo,
+  useState,
+} from 'react';
+import classNames from 'classnames';
+
+import { Person } from '../../types/Person';
+import { PersonList } from '../PersonList';
+import './Autocomplete.scss';
+
+type Props = {
+  people: Person[],
+  onSelect: (person: Person) => void,
+  delay: number,
+};
+
+export const Autocomplete: React.FC<Props> = ({
+  people,
+  onSelect,
+  delay,
+}) => {
+  const [isListVisible, setIsListVisible] = useState(false);
+  const [inputValue, setInputValue] = useState('');
+  const [debauncedInputValue, setDebauncedInputValue] = useState(inputValue);
+  const [wasThereClick, setWasThereClick] = useState(true);
+
+  const handleInputFocus = (event: React.FocusEvent<HTMLInputElement>) => {
+    if (event.target.value === '') {
+      setIsListVisible(true);
+      setWasThereClick(false);
+    }
+  };
+
+  const handleInputChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+    setInputValue(event.target.value);
+    setIsListVisible(false);
+    setWasThereClick(false);
+  };
+
+  const handleItemClick = useCallback((person: Person) => {
+    onSelect(person);
+    setInputValue(person.name);
+    setIsListVisible(false);
+    setWasThereClick(true);
+  }, []);
+
+  useEffect(() => {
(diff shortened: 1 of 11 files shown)


src/components/Autocomplete/Autocomplete.tsx
```diff
@@ -0,0 +1,101 @@
...
+};
+
+export const Autocomplete: React.FC<Props> = ({
+  people,
+  onSelect,
+  delay,
+}) => {
+  const [isListVisible, setIsListVisible] = useState(false);
+  const [inputValue, setInputValue] = useState('');
+  const [debauncedInputValue, setDebauncedInputValue] = useState(inputValue);
```
anastasiiavorobiova, line 26: There's a typo in the word 'debaunced'. It should be 'debounced'. This won't cause any errors, but it's best to fix it for readability and maintainability.

src/components/Autocomplete/Autocomplete.tsx
```diff
@@ -0,0 +1,101 @@
...
+    setIsListVisible(false);
+    setWasThereClick(false);
+  };
+
+  const handleItemClick = useCallback((person: Person) => {
+    onSelect(person);
+    setInputValue(person.name);
+    setIsListVisible(false);
+    setWasThereClick(true);
+  }, []);
```
anastasiiavorobiova, line 47: The callback functions in the useCallback hook are missing its dependency array. The current implementation may lead to bugs because it will not capture changes in the onSelect prop, and always use the initial value of onSelect and other

src/components/Autocomplete/Autocomplete.tsx
```diff
@@ -0,0 +1,101 @@
...
+    const timerId = setTimeout(() => {
+      if (inputValue !== debauncedInputValue) {
+        setDebauncedInputValue(inputValue);
+      }
+
+      setIsListVisible(true);
+    }, delay);
+
+    return () => clearTimeout(timerId);
+  }, [inputValue]);
```
anastasiiavorobiova, line 59: The useEffect hook is missing dependencies in its array. This could lead to outdated data being used inside the hook. Ensure that all variables used inside the hook are included in the dependency array, in this case 'delay' and 'debauncedInputValue' should be included.

src/components/PersonInfo/PersonInfo.tsx
```diff
@@ -0,0 +1,23 @@
...
+  person: Person,
+  onClick: (person: Person) => void,
+};
+
+export const PersonInfo: React.FC<Props> = ({ person, onClick }) => {
+  return (
+    <div
+      className="dropdown-item"
+      onClick={() => onClick(person)}
+      onKeyUp={() => {}}
```
anastasiiavorobiova, line 14: The onKeyUp event handler is currently an empty function, which doesn't provide any functionality. Consider implementing a proper event handling for keyboard events or remove this line if it's not necessary.

src/components/PersonInfo/PersonInfo.tsx
```diff
@@ -0,0 +1,23 @@
...
+      role="menuitem"
```
anastasiiavorobiova, line 15: The role 'menuitem' might not be appropriate for a non-interactive element like 'div'. Consider using a more semantic HTML tag (like 'button' or 'a') or manage the keyboard events properly.

anastasiiavorobiova (changes requested):
Great work! Added comments to fix

src/components/Autocomplete/Autocomplete.tsx
denis-kononenko: here, when I tried to add person into dependencies, I got a message about mistake that 'person can not be find' because this variable located only inside this function

PR #167, opened 2023-08-15
Solution

RodionSav:


src/App.tsx
```diff
@@ -1,14 +1,42 @@
-import React from 'react';
+import React, { useCallback, useMemo, useState } from 'react';
 import './App.scss';
+import debounce from 'lodash.debounce';
```
TarasHoliuk, line 3: the fix is described in the error - just do it
image

src/App.tsx
```diff
@@ -1,14 +1,42 @@
...
 import { peopleFromServer } from './data/people';
+import { Person } from './types/Person';
 
 export const App: React.FC = () => {
-  const { name, born, died } = peopleFromServer[0];
+  const [query, setQuery] = useState('');
+  const [appliedQuery, setAppliedQuery] = useState('');
+  const [people] = useState(peopleFromServer);
```
TarasHoliuk, line 10: this state doesn't make sense
just use `peopleFromServer`

src/App.tsx
```diff
@@ -17,40 +45,33 @@ export const App: React.FC = () => {
             type="text"
             placeholder="Enter a part of the name"
             className="input"
+            value={query}
+            onChange={handleQuery}
+            onClick={() => setHasClick(query.length > 0)}
```
TarasHoliuk, line 50: it's better to use "focus" and "blur" events to handle it
current implementation doesn't work as expected

src/App.tsx
```diff
@@ -1,14 +1,42 @@
...
+
+  const filteredPeople = useMemo(() => {
+    return people.filter(
+      person => person.name.toLowerCase().includes(appliedQuery.toLowerCase()),
+    );
+  }, [people, appliedQuery]);
+
+  const applyQuery = useCallback(debounce(setAppliedQuery, 1000), []);
+
+  const handleQuery = (event: React.ChangeEvent<HTMLInputElement>) => {
```
TarasHoliuk, line 22: try to implement this requirement:
image

hint: you have to somehow change with a delay one more state variable (like hasClick)

TarasHoliuk (changes requested):

etojeDenys (changes requested):
.

volodymyr-soltys97 (changes requested):
By default, the name of a person who is not even selected is shown, this needs to be corrected
image

src/App.tsx
IvanFesenko: Not fixed
2023-10-06_16h03_26

IvanFesenko (changes requested):
Why do I see those names?
https://github.com/mate-academy/react_autocomplete/assets/61904995/b0ce784f-89cb-42c5-995d-7ccf02261233

nazarmatsevych (approved):
👍

PR #177, opened 2023-08-21
Autocomplete react component was added

Olena-Shevchenko:
[DEMO LINK](https://Olena-Shevchenko.github.io/react_autocomplete/)

### DIFF
--- a/src/Components/Autocomplete.tsx
+++ b/src/Components/Autocomplete.tsx
@@ -1,0 +1,96 @@
+import React, {
+  useState, useMemo, useCallback, useRef,
+} from 'react';
+import { debounce } from 'lodash';
+import { Person } from '../types/Person';
+
+type Props = {
+  people: Person[];
+  onSelected: (person: Person) => void;
+};
+
+export const Autocomplete: React.FC<Props> = React.memo(({
+  people, onSelected,
+}) => {
+  const [query, setQuery] = useState('');
+  const [appliedQuery, setAppliedQuery] = useState('');
+  const [focus, setFocus] = useState(false);
+
+  const searchInput = useRef<HTMLInputElement>(null);
+
+  const applyQuery = useCallback(
+    debounce(setAppliedQuery, 1000),
+    [],
+  );
+
+  const filteredPeople = useMemo(() => {
+    return people.filter(person => person.name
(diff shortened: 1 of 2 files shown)


src/Components/Autocomplete.tsx
```diff
@@ -0,0 +1,96 @@
...
+
+      <div className="dropdown-menu" role="menu">
+        {focus
+          && (
+            <div className="dropdown-content">
+              {filteredPeople.length
+                ? filteredPeople.map(person => (
+                  <div key={person.slug} className="dropdown-item">
+                    <a
+                      href="http"
```
lerastarynets, line 74: The href attribute value 'http' is not a valid URL. You may want to use a '#'.

src/Components/Autocomplete.tsx
```diff
@@ -0,0 +1,96 @@
...
+
+    if (searchInput.current) {
+      searchInput.current.blur();
+    }
+
+    onSelected(person);
+  };
+
+  return (
+    <div className="dropdown is-active">
```
lerastarynets, line 52: The dropdown is always active because the dropdown's 'is-active' class is hard-coded. Use the focus state to conditionally add the 'is-active' class

lerastarynets (changes requested):
I see you did a hard work) Let's improve it a little bit

src/Components/Autocomplete.tsx
```diff
@@ -0,0 +1,101 @@
...
+
+      <div className="dropdown-menu" role="menu">
+        {focus
+          && (
+            <div className="dropdown-content">
+              {filteredPeople.length
+                ? filteredPeople.map(person => (
+                  <div key={person.slug} className="dropdown-item">
+                    <a
+                      href="#/"
```
lerastarynets, line 79:
```suggestion
                      href="#"
```

lerastarynets (approved):
Keep up the great work!

src/Components/Autocomplete.tsx
Olena-Shevchenko: image

Olena-Shevchenko: Linter doesn`t want me to type only "#" in href.

PR #179, opened 2023-08-22
solution

vitalii-fedusov:
[DEMO LINK](https://vitalii-fedusov.github.io/react_autocomplete/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,15 +1,66 @@
-import React from 'react';
+import debounce from 'lodash.debounce';
+import React, { useState, useCallback, useMemo } from 'react';
 import './App.scss';
+import { Person } from './types/Person';
 import { peopleFromServer } from './data/people';
+import { PeopleList } from './components/PeopleList/PeopleList';
 
 export const App: React.FC = () => {
-  const { name, born, died } = peopleFromServer[0];
+  const [selectedPerson, setSelectedPerson]
+    = useState<Person | null>(null);
+
+  const people = peopleFromServer;
+
+  const [query, setQuery] = useState('');
+  const [appliedQuery, setAppliedQuery] = useState('');
+
+  const applyQuery = useCallback(
+    debounce(setAppliedQuery, 1000),
+    [],
+  );
+
+  const [isFocused, setIsFocused] = useState(false);
+
+  const handleFocus = () => {
+    setIsFocused(true);
+  };
+
+  const handleSelected = (person: Person) => {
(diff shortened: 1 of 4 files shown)


src/App.tsx
```diff
@@ -1,55 +1,89 @@
...
+import { Person } from './types/Person';
 import { peopleFromServer } from './data/people';
+import { PeopleList } from './components/PeopleList/PeopleList';
 
 export const App: React.FC = () => {
-  const { name, born, died } = peopleFromServer[0];
+  const [selectedPerson, setSelectedPerson]
+    = useState<Person | null>(null);
+
+  const people = peopleFromServer;
```
etojeDenys, line 12: why do we need this var, if we can use peopleFromServer directly

etojeDenys (changes requested):

src/App.tsx
```diff
@@ -1,55 +1,87 @@
-import React from 'react';
+import debounce from 'lodash.debounce';
```
IvanFesenko, line 1: You did not commit updated package.json file - it might some troubles.
Also my recomendation try to create own debounce function

src/App.tsx
```diff
@@ -1,55 +1,87 @@
...
+  const [query, setQuery] = useState('');
+  const [appliedQuery, setAppliedQuery] = useState('');
+
+  const applyQuery = useCallback(
+    debounce(setAppliedQuery, 1000),
+    [],
+  );
+
+  const [isFocused, setIsFocused] = useState(false);
+
```
IvanFesenko, line 21:
```suggestion
  const [query, setQuery] = useState('');
  const [appliedQuery, setAppliedQuery] = useState('');
  const [isFocused, setIsFocused] = useState(false);
  
  const applyQuery = useCallback(
    debounce(setAppliedQuery, 1000),
    [],
  );


```

src/App.tsx
```diff
@@ -1,55 +1,87 @@
...
+  const handleQueryChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+    setQuery(event.target.value);
+    applyQuery(event.target.value);
+  };
+
+  const filteredPeople = useMemo(() => {
+    return peopleFromServer.filter(person => {
+      return person.name.toLowerCase().includes(appliedQuery.toLowerCase());
+    });
+  }, [appliedQuery, peopleFromServer]);
```
IvanFesenko, line 46: No need to filter in case query is empty

IvanFesenko (approved):

PR #189, opened 2023-08-28
solution

dseq0:
[DEMO LINK](https://dseq0.github.io/react_autocomplete/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,14 +1,41 @@
-import React from 'react';
+import React, { useState, useMemo, useCallback } from 'react';
 import './App.scss';
 import { peopleFromServer } from './data/people';
+import { Person } from './types/Person';
+import { Dropdown } from './components/Dropdown/Dropdown';
+import { debounce } from './helpers/debounce';
 
 export const App: React.FC = () => {
-  const { name, born, died } = peopleFromServer[0];
+  const [selectedPerson, setSelectedPerson] = useState<Person>();
+  const [query, setQuery] = useState('');
+  const [appliedQuery, setAppliedQuery] = useState('');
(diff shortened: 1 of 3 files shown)


src/App.tsx
```diff
@@ -1,14 +1,41 @@
...
-  const { name, born, died } = peopleFromServer[0];
+  const [selectedPerson, setSelectedPerson] = useState<Person>();
+  const [query, setQuery] = useState('');
+  const [appliedQuery, setAppliedQuery] = useState('');
+  const [isDropdownActive, setIsDropdownActive] = useState(false);
+
+  const visablePeople = useMemo(() => {
+    return peopleFromServer.filter(person => person.name
+      .toLowerCase().includes(query.toLowerCase()));
+  }, [appliedQuery]);
```
lerastarynets, line 17: The useMemo dependency should be 'query' instead of 'appliedQuery'. The 'appliedQuery' state is never used and can be removed.

src/App.tsx
```diff
@@ -17,40 +44,21 @@ export const App: React.FC = () => {
             type="text"
             placeholder="Enter a part of the name"
             className="input"
+            value={query}
+            onChange={e => {
+              handleChange(e);
+            }}
```
lerastarynets, line 50:
```suggestion
            onChange={handleChange}
```

lerastarynets (changes requested):
Nice start! Let's fix and improve your work a little bit)

sTorba24 (changes requested):
Well done! Lets improve it

text of selected option should be visible
also would be good to mark selected option(additional background, smth like that)
image

loralevitska (approved):
Well done)

PR #193, opened 2023-08-30
First commit

FallenMAD:
[DEMO LINK](https://FallenMAD.github.io/react_autocomplete/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -10,7 +10,7 @@ import { getCheckQuery } from './helper/getCheckQuery';
 export const App: React.FC = () => {
   const [query, setQuery] = useState('');
   const [people] = useState<Person[]>(peopleFromServer);
-  const [hasOpen, setHasOpen] = useState(false);
+  const [hasDropdownOpen, setHasDropdownOpen] = useState(false);
   const [apliedQuery, setApliedQuery] = useState('');
   const [selectedPerson, setSelectedPerson] = useState<Person | null>(null);
 
@@ -29,7 +29,7 @@ export const App: React.FC = () => {
     <main className="section">
       {selectedPerson ? (
         <h1 className="title">
-          {`${selectedPerson?.name} (${selectedPerson?.born} = ${selectedPerson?.died})`}
+          {`${selectedPerson.name} (${selectedPerson.born} = ${selectedPerson.died})`}
         </h1>
       ) : (
         <h1 className="title">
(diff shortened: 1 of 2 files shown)


src/PeopleList.tsx
```diff
@@ -0,0 +1,41 @@
...
+  return (
+    <div className="dropdown-menu" role="menu">
+      <div className="dropdown-content">
+        {people.length ? people.map((person) => (
+          <div
+            className="dropdown-item"
+            key={person.slug}
+          >
+            <a
+              href={person.slug}
```
Anastasiia-Svintsova, line 23: Using href for the person's slug might cause an unwanted page navigation when a suggestion is clicked. It's better to use onClick event instead of href.

src/App.tsx
```diff
@@ -1,56 +1,62 @@
...
 import { peopleFromServer } from './data/people';
+import { Person } from './types/Person';
+import { PeopleList } from './PeopleList';
+import { getCheckQuery } from './helper/getCheckQuery';
 
 export const App: React.FC = () => {
-  const { name, born, died } = peopleFromServer[0];
+  const [query, setQuery] = useState('');
+  const [people] = useState<Person[]>(peopleFromServer);
+  const [hasOpen, setHasOpen] = useState(false);
```
Anastasiia-Svintsova, line 13: The hasOpen state variable name is not descriptive. Consider renaming to isDropdownOpen for better readability.

src/App.tsx
```diff
@@ -1,56 +1,62 @@
...
     <main className="section">
-      <h1 className="title">
-        {`${name} (${born} = ${died})`}
-      </h1>
-
-      <div className="dropdown is-active">
-        <div className="dropdown-trigger">
+      {selectedPerson ? (
+        <h1 className="title">
+          {`${selectedPerson?.name} (${selectedPerson?.born} = ${selectedPerson?.died})`}
```
Anastasiia-Svintsova, line 32: You've already checked if selectedPerson exists. Why do you need these `?`

src/PeopleList.tsx
```diff
@@ -0,0 +1,41 @@
...
+              className="has-text-link link"
+              onMouseDown={() => {
+                onSelected(person);
+                setQuery(person.name);
+              }}
+            >
+              {person.name}
```
Anastasiia-Svintsova, line 30: Don't forget about objects destructuring

src/PeopleList.tsx
```diff
@@ -0,0 +1,41 @@
...
+            className="dropdown-item"
+            key={person.slug}
+          >
+            <a
+              href={person.slug}
+              className="has-text-link link"
+              onMouseDown={() => {
+                onSelected(person);
+                setQuery(person.name);
+              }}
```
Anastasiia-Svintsova, line 28: Move it to a separate function to improve the code readability and keep your JSX code clean

Anastasiia-Svintsova (changes requested):
- The user should be able to interact with autocomplete with the keyboard (I mean arrows to switch between options and enter to choose the option). Now it's not possible

src/App.tsx
FallenMAD: my bad

mykhalenych (approved):

PR #203, opened 2023-09-05
Solution

antonina-klishch:
[LINK](https://antonina-klishch.github.io/react_autocomplete/)

### DIFF
--- a/src/components/PeopleList.tsx
+++ b/src/components/PeopleList.tsx
@@ -31,7 +31,7 @@ export const PeopleList: React.FC<Props> = ({
       </button>
     ))}
 
-    {people.length === 0 && (
+    {!people.length && (
       <span className="has-text-danger">
         No matching suggestions
       </span>


src/App.tsx
```diff
@@ -1,54 +1,79 @@
-import React from 'react';
+import React, {
+  useCallback,
+  useMemo,
+  useState,
+} from 'react';
+import debounce from 'lodash.debounce';
```
VitaliyBondarenko1982, line 6: for fix such linter issue on github, commit and push package.json file
Screenshot from 2023-09-05 16-37-52

VitaliyBondarenko1982 (changes requested):
Good job.
Needs some improvements.
- Dropdown shuldn't be visible, when input is not in focus.
chrome-capture-2023-8-5 (1)

src/App.tsx
etojeDenys: still not fixed, you need to write `git add package.json`
image

etojeDenys (changes requested):
let's also save the name in query input after selecting
image

src/components/PeopleList.tsx
```diff
@@ -0,0 +1,40 @@
...
+            'has-text-danger': person.sex === 'f',
+            'has-text-link': person.sex === 'm',
+          })}
+        >
+          {person.name}
+        </p>
+      </button>
+    ))}
+
+    {people.length === 0 && (
```
lerastarynets, line 34:
```suggestion
    {!people.length && (
```

lerastarynets (approved):
Great job overall! Left one tiny suggestion for the improvement

PR #206, opened 2023-09-07
Solved

silcki:
https://silcki.github.io/react_autocomplete/


anastasiia-tilikina (changes requested):
Good enough. 
let's implement this missing requirement 
image

Also I belive that you should hide dropdown on blur

silcki:
@anastasiia-tilikina up

PR #208, opened 2023-09-07
Implement ability to choose user by name

Koliras:
[DEMO LINK](https://Koliras.github.io/react_autocomplete/)


src/components/DropdownMenu/DropdownMenu.tsx
```diff
@@ -0,0 +1,52 @@
...
+export const DropdownMenu: React.FC<Props> = React.memo(
+  ({ people, onSelected }) => {
+    return (
+      <div
+        className="dropdown-menu"
+        role="menu"
+        id="dropdown-menu"
+      >
+        <div className="dropdown-content">
+          {people.length === 0
```
davinnchii, line 19: It's better to use just people.length, since u use it with ternary operator as Max stated in Slack.
image

src/App.tsx
```diff
@@ -1,14 +1,48 @@
...
+  const filteredPeople = useMemo(() => {
+    return peopleFromServer.filter(
+      person => person.name.toLowerCase().includes(appliedQuery.toLowerCase()),
+    );
+  }, [appliedQuery]);
+
+  const applyQuery = useCallback(
+    debounce(setAppliedQuery, 1000),
+    [],
+  );
```
davinnchii, line 24: Also you could create a hook-file useDebounce and call useCallback there. In this case you would just call ur new hook and pass function there. It would look like this:
```js
const applyQuery= useDebounce(setAppliedQuery, 1000);
```
Here you can find more about useDebounce
https://www.developerway.com/posts/debouncing-in-react

davinnchii (approved):
GJ, but fix comment about ternary operator in DropdownMenu.tsx
Shouldn't it clear the search field after u select a user? I'm not sure, but think it should.
image

src/App.tsx
alexander-ignatow, line 24: Unnecessary `useCallback` and magic value of `1000`

src/App.tsx
```diff
@@ -1,14 +1,48 @@
...
+    setAppliedQuery(person.name);
+    setIsListVisible(false);
+  }, []);
 
   return (
     <main className="section">
       <h1 className="title">
-        {`${name} (${born} = ${died})`}
+        {selectedUser
+          ? `${selectedUser?.name} (${selectedUser?.born} = ${selectedUser?.died})`
```
alexander-ignatow, line 44:
```suggestion
          ? `${selectedUser?.name} (${selectedUser?.born} - ${selectedUser?.died})`
```

src/App.tsx
```diff
@@ -17,40 +51,21 @@ export const App: React.FC = () => {
             type="text"
             placeholder="Enter a part of the name"
             className="input"
+            aria-haspopup="true"
+            aria-controls="dropdown-menu"
+            value={query}
+            onChange={handleInputChange}
+            onFocus={() => setIsListVisible(true)}
```
alexander-ignatow, line 58: handle `onBlur` event to hide the dropdown when input loses focus

src/components/DropdownMenu/DropdownMenu.tsx
```diff
@@ -0,0 +1,52 @@
+import React from 'react';
+import classNames from 'classnames';
+import { Person } from '../../types/Person';
+
+type Props = {
+  people: Person[],
+  onSelected: (person: Person) => void,
+};
+
+export const DropdownMenu: React.FC<Props> = React.memo(
```
alexander-ignatow, line 10: You should use React.memo only when you really have some problems and need to reduce component re-renders by parent

src/components/DropdownMenu/DropdownMenu.tsx
```diff
@@ -0,0 +1,52 @@
...
+            ? (
+              <div className="dropdown-item" style={{ cursor: 'pointer' }}>
+                <p>
+                  No matching suggestions
+                </p>
+              </div>
+            ) : (
+              people.map(person => (
+
+                /* eslint-disable */
```
alexander-ignatow, line 29: Please enable eslint and fix a11y errors :)

alexander-ignatow (changes requested):
Good, but some changes needed

I suggest you move input (and corresponding statuses) inside `DropdownMenu` and expand component props per task:
> The delay should be customizable via props;

src/App.tsx
```diff
@@ -1,57 +1,23 @@
...
 export const App: React.FC = () => {
-  const { name, born, died } = peopleFromServer[0];
+  const [selectedUser, setSelectedUser] = useState<Person | null>(null);
 
   return (
     <main className="section">
       <h1 className="title">
-        {`${name} (${born} = ${died})`}
+        {selectedUser
+          ? `${selectedUser?.name} (${selectedUser?.born} - ${selectedUser?.died})`
```
roman-mirzoian, line 13: Maybe you forgot to redeploy the application?
image

src/components/DropdownMenu/DropdownMenu.tsx
```diff
@@ -0,0 +1,97 @@
...
+            {!filteredPeople.length
+              ? (
+                <div className="dropdown-item" style={{ cursor: 'pointer' }}>
+                  <p>
+                    No matching suggestions
+                  </p>
+                </div>
+              ) : (
+                filteredPeople.map(person => (
+
```
roman-mirzoian, line 73:
```suggestion
```

src/components/DropdownMenu/DropdownMenu.tsx
```diff
@@ -0,0 +1,97 @@
...
+                    className="dropdown-item"
+                    key={person.slug}
+                    onMouseDown={() => handleUserSelection(person)}
+                    style={{ cursor: 'pointer' }}
+                    aria-hidden="true"
+                  >
+                    <p
+                      className={classNames({
+                        'has-text-link': person.sex === 'm',
+                        'has-text-danger': person.sex === 'f',
```
roman-mirzoian, line 84: Avoid using "magic values".

roman-mirzoian (changes requested):
Polish your code, redeploy the app and fix the tests, because not all GitHub tests are successful.

src/components/DropdownMenu/DropdownMenu.tsx
```diff
@@ -0,0 +1,99 @@
...
+                  <p>
+                    No matching suggestions
+                  </p>
+                </div>
+              ) : (
+                filteredPeople.map(person => (
+                  <div
+                    className="dropdown-item"
+                    key={person.slug}
+                    onMouseDown={() => handleUserSelection(person)}
```
alexander-ignatow, line 79: `onMouseDown` is an incorrect event here - consider handling `onClick` instead

src/components/DropdownMenu/DropdownMenu.tsx
```diff
@@ -0,0 +1,99 @@
...
+    setAppliedQuery(person.name);
+    setIsListVisible(false);
+  }, []);
+
+  const handleInputChange = useCallback(
+    (event: React.ChangeEvent<HTMLInputElement>) => {
+      setQuery(event.target.value);
+      applyQuery(event.target.value);
+    }, [],
+  );
```
alexander-ignatow, line 42: These `useCallback` hoks does not have any effect here

alexander-ignatow (changes requested):
God progress!

You are almost done - just few more minor items to fix

src/components/DropdownMenu/DropdownMenu.tsx
Koliras: I cannot do this. I've talked about this with one of the mentors.
On the line 57 there is onBlur which sets the visibility of the dropdown menu to false, and because of that after click on the name dropdown firstly disappears and then the user can't choose the person. So to fix this I used onMouseDown so this command will activate earlier than onBlur

BudnikOleksii (approved):
Good work 👍

PR #209, opened 2023-09-07
Implemented logic for choosing user from dropdown

davinnchii:
[DEMO LINK](https://davinnchii.github.io/react_autocomplete/)

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,4 +20,4 @@ that will suggest people matching an entered text.
 
 - Implement a solution following the [React task guideline](https://github.com/mate-academy/react_task-guideline#react-tasks-guideline).
 - Use the [React TypeScript cheat sheet](https://mate-academy.github.io/fe-program/js/extra/react-typescript).
-- Replace `<your_account>` with your Github username in the [DEMO LINK](https://<your_account>.github.io/react_autocomplete/) and add it to the PR description.
+- Replace `<your_account>` with your Github username in the [DEMO LINK](https://davinnchii.github.io/react_autocomplete/) and add it to the PR description.
--- a/src/components/DropDown/index.tsx
+++ b/src/components/DropDown/index.tsx
@@ -1,0 +1,71 @@
+import cn from 'classnames';
+import React, { ChangeEvent } from 'react';
+import { Person } from '../../types/Person';
+
+type Props = {
+  onSelected: (person: Person) => void,
+  query: string,
+  isFocused: boolean,
+  setIsFocused: (isFocused: boolean) => void,
+  people: Person[] | undefined,
+  onSearch: (event: ChangeEvent<HTMLInputElement>) => void,
+};
+
+export const DropDown: React.FC<Props> = ({
+  people,
+  onSelected,
+  isFocused,
+  onSearch,
+  setIsFocused,
+  query,
+}) => {
+  return (
+    <div className={cn('dropdown',
+      { 'is-active': isFocused })}
+    >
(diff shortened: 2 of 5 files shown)


README.md
```diff
@@ -20,4 +20,4 @@ that will suggest people matching an entered text.
 
 - Implement a solution following the [React task guideline](https://github.com/mate-academy/react_task-guideline#react-tasks-guideline).
 - Use the [React TypeScript cheat sheet](https://mate-academy.github.io/fe-program/js/extra/react-typescript).
-- Replace `<your_account>` with your Github username in the [DEMO LINK](https://<your_account>.github.io/react_autocomplete/) and add it to the PR description.
+- Replace `<your_account>` with your Github username in the [DEMO LINK](https://davinnchii.github.io/react_autocomplete/) and add it to the PR description.
```
DenisChakhalian, line 23: link doesn't work 😅

src/components/DropDown/index.tsx
```diff
@@ -0,0 +1,71 @@
+import cn from 'classnames';
+import React, { ChangeEvent } from 'react';
+import { Person } from '../../types/Person';
```
DenisChakhalian, line 3: it is better to write the full name of the library that is used as a variable (instead of cn - classnames)

DenisChakhalian (approved):
looks great!

src/components/DropDown/index.tsx
```diff
@@ -0,0 +1,71 @@
...
+
+type Props = {
+  onSelected: (person: Person) => void,
+  query: string,
+  isFocused: boolean,
+  setIsFocused: (isFocused: boolean) => void,
+  people: Person[] | undefined,
```
BudnikOleksii, line 10: You don't need to render the `DropDown` component if you don't have any people.
Also, if in typescript you expect that value can be undefined, you can do it like this `people?: Person[]`

src/useDebounce.ts
```diff
@@ -0,0 +1,37 @@
...
+    return debounce((...args: Parameters<T>) => {
+      callbackRef.current?.(...args);
+    }, delay);
+  }, [delay]);
+
+  useEffect(() => {
+    return () => {
+      debouncedCallback.cancel();
+    };
+  }, [debouncedCallback]);
```
BudnikOleksii, line 29: Looks too complicated, you can use `use-debounce` library with ready to use hook:
https://www.npmjs.com/package/use-debounce

BudnikOleksii (changes requested):
Good start 👍 
Left some comments, check them, and fix in accordance. Also, the preview link is broken, deploy your solution so a reviewer can check if it works as expected

src/components/DropDown/index.tsx
polinavafik: It is actually okay to use cn instead of classnames, because a lot of developers do that (including misha), so that is a kind of a thing that is agreed on a project
+ it saves up "space" in the line of code 🤗

src/components/DropDownItem/index.tsx
```diff
@@ -0,0 +1,35 @@
...
+  return (
+    <>
+      <button
+        type="button"
+        onClick={() => onSelected(person)}
+      >
+        <p className={cn({
+          'has-text-link': person.sex === 'm',
+          'has-text-danger': person.sex === 'f',
+        })}
```
roman-mirzoian, line 25: Avoid using "magic values".

roman-mirzoian (approved):
Well done 👍

PR #214, opened 2023-09-08
Task solution - implemented Autocomplete function

yuliiaant:
[DEMO LINK](https://yuliiaant.github.io/react_autocomplete/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,56 +1,88 @@
-import React from 'react';
+import React, { useCallback, useMemo, useState } from 'react';
+import { debounce } from 'lodash';
 import './App.scss';
 import { peopleFromServer } from './data/people';
+import { Person } from './types/Person';
+import { DropdownMenu } from './DropdownMenu';
+import { DropdownInput } from './DropdownInput';
+
+const search = (people: Person[], query: string) => {
+  const preparedQuery = query.trim().toLowerCase();
+
+  return people
+    .filter(person => person.name.toLowerCase().includes(preparedQuery));
+};
 
 export const App: React.FC = () => {
-  const { name, born, died } = peopleFromServer[0];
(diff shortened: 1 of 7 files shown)


src/App.tsx
```diff
@@ -1,56 +1,88 @@
...
-            <div className="dropdown-item">
-              <p className="has-text-link">Pieter Haverbeke</p>
-            </div>
+  const onSelect = useCallback(
+    (person: Person) => {
+      setSelectedPerson(person);
+      setIsListShown(false);
+    },
+    [debouncedQuery],
+  );
```
DenisChakhalian, line 37: task condition :
image
in your situation is not saved in input

src/App.tsx
```diff
@@ -1,56 +1,88 @@
...
 
-            <div className="dropdown-item">
-              <p className="has-text-link">Pieter Bernard Haverbeke</p>
-            </div>
+  const handleQueryChange = useCallback((newQuery: string) => {
+    setQuery(newQuery);
+    applyDebounceQuery(newQuery);
+  }, [debouncedQuery]);
```
DenisChakhalian, line 43: task condition :
image

DenisChakhalian (approved):
good job! but there are some things that can be fixed if you want

alexander-ignatow (changes requested):
You have good solution, but some refactor is needed

1. You are overusing useMemo/useCallback/React.memo functionality. In this very case, it gives you nearly zero but makes your code more complex and even degrades performance (to some extent). I suggest you remove all memorization.
2. You did a good job separating all into components - now let's take one last step and create an `Autocomplete` component to be an `entry point` and use it in the `App`

BudnikOleksii (approved):
Nicely done 🔥

PR #217, opened 2023-09-08
create a search dropdown and add its fuctional

rostyslav48:
[DEMO LINK](https://rostyslav48.github.io/react_autocomplete/)

The test is crashing because of not installed library I suppose

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -17,40 +17,59 @@ function filterPeopleByQuery(peopleToFilter: Person[], query: string) {
 export const App: React.FC = () => {
   const [query, setQuery] = useState('');
   const [appliedQuery, setAppliedQuery] = useState('');
+  const [isDropdownVisible, setIsDropdownVisible] = useState(false);
+
+  const applyQuery
+  = useCallback(debounce((value: string) => {
+    setIsDropdownVisible(true);
+
+    return setAppliedQuery(value);
+  }, 1000), [appliedQuery]);
+
+  const handleQueryChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+    setIsDropdownVisible(false);
+    setQuery(event.target.value);
+    applyQuery(event.target.value.trim());
+  };
 
   const filteredPeople: Person[] = useMemo(() => {
     return filterPeopleByQuery(peopleFromServer, appliedQuery);
   }, [appliedQuery]);
 
-  const applyQuery
-  = useCallback(debounce(setAppliedQuery, 1000), [appliedQuery]);
+  const [selectedPerson, setSelectedPerson]
+    = useState<Person | null>(null);
 
-  const handleQueryChange = (event: React.ChangeEvent<HTMLInputElement>) => {
-    setQuery(event.target.value);
-    applyQuery(event.target.value.trim());
+  const onSelect = (person: Person) => {
+    setSelectedPerson(person);
+    setQuery(person.name);
+    setAppliedQuery(person.name);
+    setIsDropdownVisible(false);
   };
 
-  const [selectedPerson, setSelectedPerson]
-    = useState<Person>(filteredPeople[0]);
-
-  const { name, born, died } = selectedPerson;
+  const titleMessage = selectedPerson
+    ? `${selectedPerson.name} (${selectedPerson.born} - ${selectedPerson.died})`
+    : 'No selected person';
 
   return (
     <main className="section">
       <h1 className="title">
-        {`${name} (${born} - ${died})`}
+        {titleMessage}
       </h1>
 
       <div className="dropdown is-active">
         <SearchBar
+          setIsDropdownVisible={setIsDropdownVisible}
           handleQueryChange={handleQueryChange}
           query={query}
         />
 
-        <DropdownList
-          filteredPeople={filteredPeople}
-          setSelectedPerson={setSelectedPerson}
-        />
+        {isDropdownVisible && (
+          <DropdownList
+            filteredPeople={filteredPeople}
+            onSelect={onSelect}
+          />
+        )}
+
       </div>
     </main>
   );
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,50 +1,13 @@
-import React, { useCallback, useMemo, useState } from 'react';
-import { debounce } from 'lodash';
+import React, { useState } from 'react';
 
 import './App.scss';
-import { peopleFromServer } from './data/people';
+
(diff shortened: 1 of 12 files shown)


Wesses:
lgtm

Wesses (approved):
lgtm

src/App.tsx
```diff
@@ -1,56 +1,75 @@
...
+  const [query, setQuery] = useState('');
+  const [appliedQuery, setAppliedQuery] = useState('');
+  const [isDropdownVisible, setIsDropdownVisible] = useState(false);
+
+  const applyQuery
+  = useCallback(debounce((value: string) => {
+    setIsDropdownVisible(true);
+
+    return setAppliedQuery(value);
+  }, 1000), [appliedQuery]);
```
alexander-ignatow, line 27: Per task:
> the delay should be customizable via props;

In addition - `1000` is a magic number

alexander-ignatow (changes requested):
Good 👍 

But one refactor is needed - I'd suggest you to create a component `Autocomplete` and incapsulate your sub-components within it

Also you need to fix one issue in the code

src/App.tsx
```diff
@@ -1,57 +1,28 @@
...
-            <div className="dropdown-item">
-              <p className="has-text-danger">Elisabeth Hercke</p>
-            </div>
-          </div>
-        </div>
-      </div>
+      <Autocomplete
+        setSelectedPerson={setSelectedPerson}
+        delay={1000}
+      />
```
alexander-ignatow, line 25: Now looks much cleaner 👍

src/components/Autocomplete/Autocomplete.tsx
```diff
@@ -0,0 +1,68 @@
...
+  delay: number,
+};
+
+export const Autocomplete: React.FC<Props> = ({ setSelectedPerson, delay }) => {
+  const [query, setQuery] = useState('');
+  const [appliedQuery, setAppliedQuery] = useState('');
+  const [isDropdownVisible, setIsDropdownVisible] = useState(false);
+
+  const applyQuery
+  = useCallback(debounce((value: string) => {
```
alexander-ignatow, line 28: This `useCallback` looks unnecessary

src/components/DropdownPerson/DropdownPerson.tsx
```diff
@@ -0,0 +1,38 @@
...
+  onSelect,
+}) => {
+  const isMale = (sex: string) => sex === 'm';
+
+  return (
+    <div
+      className="dropdown-item"
+    >
+      <button
+        type="button"
```
alexander-ignatow, line 22: `button` is a good pick here 👍

src/components/SearchBar/SearchBar.tsx
```diff
@@ -0,0 +1,24 @@
...
+};
+
+export const SearchBar: React.FC<Props> = ({
+  query,
+  handleQueryChange,
+  setIsDropdownVisible,
+}) => (
+  <div className="dropdown-trigger">
+    <input
+      onFocus={() => setIsDropdownVisible(true)}
```
alexander-ignatow, line 16: To make it perfect - it is best to handle `onBlur` - to hide the suggestion, when component looses the focus

alexander-ignatow (changes requested):
Almost perfect!

Just two last issues remain

src/components/Autocomplete/Autocomplete.tsx
rostyslav48: It is needed to prevent extra rerendering (the callback creates every time the query changes)

src/components/SearchBar/SearchBar.tsx
rostyslav48: I have added it, however to make it work we need to add some delay before the dropdown collapses. It is needed to have the opportunity to handle the button click.

src/components/Autocomplete/Autocomplete.tsx
```diff
@@ -0,0 +1,68 @@
...
+        query={query}
+      />
+
+      {isDropdownVisible && (
+        <DropdownList
+          filteredPeople={filteredPeople}
+          onSelect={onSelect}
+        />
+      )}
+
```
roman-mirzoian, line 65:
```suggestion
```

src/components/DropdownPerson/DropdownPerson.tsx
```diff
@@ -0,0 +1,38 @@
...
+type Props = {
+  person: Person,
+  onSelect: (person: Person) => void,
+};
+
+export const DropdownPerson: React.FC<Props> = ({
+  person,
+  onSelect,
+}) => {
+  const isMale = (sex: string) => sex === 'm';
```
roman-mirzoian, line 15: This function can be taken outside the component.
Also, avoid using magic values.

roman-mirzoian (changes requested):
Good work, but currently not all tests on github pass.

src/components/Autocomplete/Autocomplete.tsx
rostyslav48: This lines linter put automatically

src/components/DropdownPerson/DropdownPerson.tsx
rostyslav48: There is no tests for this task
Знімок екрана 2023-09-15 о 09 15 34

roman-mirzoian (changes requested):
Github tests glow red - so something went wrong:
image

src/components/Autocomplete/Autocomplete.tsx
alexander-ignatow: `useCallback` only does one thing - it stores the current function object, so it survives renders, unless the dependencies change.

It is useful in basically two scenarios:
- It is the dependency for another hook (like useMemo, useCallback, useEffect) 
- It is passed as a property to the memorized component

In your case, you just use `applyQuery` in another callback function, so usage of the useCallback has no benefits. Even more - any memorization without clear gain is harmful, as it adds code complexity, degrades performance, and adds a perfect space for hard-to-trace bugs.

src/components/SearchBar/SearchBar.tsx
alexander-ignatow: Looks like a hack :)

Ok, though I believe there are other ways to fix this, IMO your solution is not bad, so let's keep it 👍

BudnikOleksii (approved):
👍

PR #219, opened 2023-09-09
Added React_autocomplete

vserhiichuk:
[DEMO LINK](https://vserhiichuk.github.io/react_autocomplete/)


src/App.tsx
```diff
@@ -1,57 +1,55 @@
...
-            </div>
-
-            <div className="dropdown-item">
-              <p className="has-text-link">Pieter Bernard Haverbeke</p>
-            </div>
-
-            <div className="dropdown-item">
-              <p className="has-text-link">Pieter Antone Haverbeke</p>
-            </div>
+  const serverQuery = useCallback(
```
davinnchii, line 13: I think better name here would be like applyQuery, if we debounce setting AppliedQuery

src/App.tsx
```diff
@@ -1,57 +1,55 @@
...
-              <p className="has-text-danger">Elisabeth Hercke</p>
-            </div>
-          </div>
-        </div>
-      </div>
+  return (
+    <main className="section">
+      {selectedPeson ? (
+        <h1 className="title">
+          {`${selectedPeson.name} (${selectedPeson.born} = ${selectedPeson.died})`}
```
davinnchii, line 38: You can move it to some helper function, which will return you this string.

src/components/ListPeople/ListPeople.tsx
```diff
@@ -0,0 +1,67 @@
...
+                      className={cn('dropdown-item', {
+                        'has-text-link': person.sex === 'm',
+                        'has-text-danger': person.sex === 'f',
+                      })}
+                      href={`${person.slug}`}
+                      onClick={(event) => {
+                        event.preventDefault();
+
+                        onSelected(person);
+                      }}
```
davinnchii, line 52: I think we can move it to handler or optionally, we can use button instead of a in this case, so we won't have to preventDefault

src/App.tsx
```diff
@@ -1,57 +1,55 @@
...
+import debounce from 'lodash.debounce';
 import { peopleFromServer } from './data/people';
+import { Person } from './types/Person';
+import { ListPeople } from './components/ListPeople/ListPeople';
 
 export const App: React.FC = () => {
-  const { name, born, died } = peopleFromServer[0];
+  const [query, setQuery] = useState('');
+  const [appliedQuery, setAppliedQuery] = useState('');
+  const [selectedPeson, setSelectedPeson] = useState<Person | null>(null);
```
davinnchii, line 11: U missed 'r' in name selectedPerson

davinnchii (approved):
Gj, you can simplify your code if you want

src/components/ListPeople/ListPeople.tsx
```diff
@@ -0,0 +1,67 @@
...
+import { Person } from '../../types/Person';
+
+interface Props {
+  people: Person[]
+  query: string,
+  onQuery?: (event: React.ChangeEvent<HTMLInputElement>) => void,
+  onSelected?: (person: Person) => void
+}
+
+export const ListPeople: React.FC<Props> = React.memo(
```
alexander-ignatow, line 12: You made the correct separation, but this component needs a refactor

1. `Autocomplete` or `PeopleAutocomplete` is a better name for the component
2. Please implement all properties, as required by the task
3. What is the reason for storing `query` outside the component? You have the input inside, so you should incapsulate this state inside as well

src/components/ListPeople/ListPeople.tsx
```diff
@@ -0,0 +1,67 @@
...
+              ? (
+                <div className="dropdown-content">
+                  {people.map(person => (
+                    <a
+                      key={person.slug}
+                      className={cn('dropdown-item', {
+                        'has-text-link': person.sex === 'm',
+                        'has-text-danger': person.sex === 'f',
+                      })}
+                      href={`${person.slug}`}
```
alexander-ignatow, line 47: Not bad idea of providing some meaningful `href` for `a` tag 👍

alexander-ignatow (changes requested):
Good, but there are some issues to fix

Also, you have a bug, as the filtering works wrong
image

image

src/App.tsx
```diff
@@ -1,57 +1,31 @@
...
+import { Person } from './types/Person';
+import { Autocomplete } from './components/Autocomplete/Autocomplete';
+
+const DEBOUNCE_DELAY = 1000;
 
 export const App: React.FC = () => {
-  const { name, born, died } = peopleFromServer[0];
+  const [selectedPerson, setSelectedPerson] = useState<Person | null>(null);
+
+  function formatPerson(person: Person | null) {
```
roman-mirzoian, line 11:
```suggestion
  function getPersonInfo(person: Person | null) {
```

roman-mirzoian, line 11: Such a function can be taken outside the component.

src/components/Autocomplete/Autocomplete.tsx
```diff
@@ -0,0 +1,84 @@
...
+import { peopleFromServer } from '../../data/people';
+
+interface Props {
+  onSelected?: (person: Person) => void,
+  delay: number
+}
+
+export const Autocomplete: React.FC<Props> = React.memo(
+  ({
+    onSelected = () => { },
```
roman-mirzoian, line 15: Why set an empty callback by default?

src/components/Autocomplete/Autocomplete.tsx
```diff
@@ -0,0 +1,84 @@
...
+          <div className="dropdown-menu" role="menu">
+            {filteredPeople.length
+              ? (
+                <div className="dropdown-content">
+                  {filteredPeople.map(person => (
+                    <a
+                      key={person.slug}
+                      className={cn('dropdown-item', {
+                        'has-text-link': person.sex === 'm',
+                        'has-text-danger': person.sex === 'f',
```
roman-mirzoian, line 66: Avoid using "magic values".

roman-mirzoian (changes requested):

src/components/Autocomplete/Autocomplete.tsx
```diff
@@ -0,0 +1,87 @@
...
+import cn from 'classnames';
+
+import { Person } from '../../types/Person';
+import { peopleFromServer } from '../../data/people';
+
+const MALE_SEX = 'm';
+const FEMALE_SEX = 'f';
+
+interface Props {
+  onSelected?: (person: Person) => void,
```
BudnikOleksii, line 12:
```suggestion
  onSelected: (person: Person) => void,
```

BudnikOleksii (approved):
Good work 👍

PR #222, opened 2023-09-10
add task solution

cty2802:
[DEMO LINK](https://cty2802.github.io/react_autocomplete/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,56 +1,86 @@
-import React from 'react';
+import React, {
+  useCallback,
+  useMemo,
+  useRef,
+  useState,
+} from 'react';
 import './App.scss';
+import debounce from 'lodash.debounce';
 import { peopleFromServer } from './data/people';
+import { SearchBar } from './component/SearchBar';
+
+const preparedPeople = [...peopleFromServer];
 
 export const App: React.FC = () => {
-  const { name, born, died } = peopleFromServer[0];
+  const [people] = useState(preparedPeople);
+  const [filteredPeople, setFilteredPeople] = useState(people);
+  const [query, setQuery] = useState('');
+  const [appliedQuery, setAppliedQuery] = useState('');
+  const [focused, setFocused] = useState(false);
+
+  const inputField = useRef<HTMLInputElement>(null);
+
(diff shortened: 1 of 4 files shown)


Vannnkof (approved):
Code looks good, but demo link works strange

src/App.tsx
```diff
@@ -1,56 +1,86 @@
...
+  const [query, setQuery] = useState('');
+  const [appliedQuery, setAppliedQuery] = useState('');
+  const [focused, setFocused] = useState(false);
 
-  return (
-    <main className="section">
-      <h1 className="title">
-        {`${name} (${born} = ${died})`}
-      </h1>
+  const inputField = useRef<HTMLInputElement>(null);
```
alexander-ignatow, line 21: Looks, that you are not using this ref

src/App.tsx
```diff
@@ -1,56 +1,86 @@
...
 
-        <div className="dropdown-menu" role="menu">
-          <div className="dropdown-content">
-            <div className="dropdown-item">
-              <p className="has-text-link">Pieter Haverbeke</p>
-            </div>
+  const applyQuery = useCallback(
+    debounce(setAppliedQuery, 500),
+    [],
+  );
```
alexander-ignatow, line 34: this `useCalback` is useless

src/App.tsx
```diff
@@ -1,56 +1,86 @@
...
-      <div className="dropdown is-active">
-        <div className="dropdown-trigger">
-          <input
-            type="text"
-            placeholder="Enter a part of the name"
-            className="input"
-          />
-        </div>
+  const { name, born, died } = filteredPeople?.length
+    ? filteredPeople[0] : { name: '', born: '', died: '' };
```
alexander-ignatow, line 24: This is wrong - you should be able to select the person - not just pick one from the top

alexander-ignatow (changes requested):
Hey!
The main problem of our solution is that you probably got wrong the idea, how autocomplete should work

Please review your solution and fix the core logic

As an inspiration (but not a direct example) you may check out MUI's autocomplete
https://mui.com/material-ui/react-autocomplete/

PR #226, opened 2023-09-11
add task soltion

Orpheus29:
[DEMO LINK](https://Orpheus29.github.io/react_autocomplete/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,57 +1,38 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
 import { peopleFromServer } from './data/people';
+import { Autocomplete } from './components/Autocomplete';
+import { PeoplesNames } from './types/Person';
+
+const DELAY = 1000;
+
+const peoplesNames: PeoplesNames[] = peopleFromServer.map(({
+  name,
+  slug,
+  sex,
+}) => ({
+  name,
+  slug,
+  sex,
+}));
 
 export const App: React.FC = () => {
-  const { name, born, died } = peopleFromServer[0];
+  const [selectedPersonId, setSelectedPersonId] = useState<string>('');
+  const selectedPerson = peopleFromServer
+    .find(person => person.slug === selectedPersonId);
 
   return (
     <main className="section">
       <h1 className="title">
-        {`${name} (${born} = ${died})`}
+        {selectedPerson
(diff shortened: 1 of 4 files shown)


src/App.tsx
```diff
@@ -1,57 +1,38 @@
...
+  name,
+  slug,
+  sex,
+}));
 
 export const App: React.FC = () => {
-  const { name, born, died } = peopleFromServer[0];
+  const [selectedPersonId, setSelectedPersonId] = useState<string>('');
+  const selectedPerson = peopleFromServer
+    .find(person => person.slug === selectedPersonId);
```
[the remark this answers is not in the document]
Orpheus29: I also don't know what it's better, to pass the whole bunch to the component or to pick necessary properties beforehand.

src/components/Autocomplete.tsx
```diff
@@ -0,0 +1,108 @@
...
+    debounce(setProcessedQuery, delay),
+    [],
+  );
+
+  const handleQueryChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+    setInputQuery(event.target.value);
+    handleDelayedChange(event.target.value);
+  };
+
+  useMemo(() => {
```
BudnikOleksii, line 38: You `useMemo` like `useEffect` it doesn't make sense.
[Docs](https://react.dev/reference/react/useMemo)

If you want to use useMemo here, it should look like this, and then render only `filteredSuggestions`

```javascript
const filteredSuggestions = useMemo(
    () => peoplesNames
      .filter((person) => person.name.toLowerCase().includes(processedQuery.toLowerCase())),
    [processedQuery]
  );
```

src/components/Autocomplete.tsx
```diff
@@ -0,0 +1,108 @@
...
+
+      <div
+        className="dropdown-menu"
+        role="menu"
+      >
+        {!suggestions.length
+          ? (
+            <div className="dropdown-item">
+              <p className="has-text-danger">No matching suggestions</p>
+              {onSelected('')}
```
BudnikOleksii, line 74: I have no idea why you do it, it's a bad idea to change the state like this

BudnikOleksii (changes requested):
Good work 👍 
Left some comments, check them

BudnikOleksii (approved):
Good work 🚀

PR #228, opened 2023-09-11
Solution

Oleh0293:
[DEMO LINK](https://Oleh0293.github.io/react_autocomplete/)


src/App.tsx
```diff
@@ -1,57 +1,83 @@
...
+    return `${name} (${born} - ${died})`;
+  }
+
+  return 'No selected person';
+}
 
 export const App: React.FC = () => {
-  const { name, born, died } = peopleFromServer[0];
+  const [selectedPerson, setSelectedPerson] = useState<Person | null>(null);
+  const [people, setPeople] = useState<Person[]>(peopleFromServer);
```
Denys14Samoilenko, line 20: i don't understand why you need this state.

src/components/DropDownContent.tsx
```diff
@@ -0,0 +1,65 @@
...
+          onClick={() => setIsFocused(!isFocused)}
+          onChange={onSearch}
+        />
+      </div>
+
+      <div className="dropdown-menu" role="menu">
+        {people?.length && query
+          ? (
+            <div className="dropdown-content">
+              {people.map((person, index) => {
```
Denys14Samoilenko, line 42: You need to map filteredPeople

Denys14Samoilenko (changes requested):
Nice work! 
But now if demo link work rigth, when you type something in input your app doesn't filter your array and return your initial array.

PR #229, opened 2023-09-11
added logic for autocmplete task

irahos:
[DEMO LINK](https://irahos.github.io/react_autocomplete/)


src/App.tsx
```diff
@@ -1,56 +1,76 @@
...
+    });
+  }, [appliedQuery, peopleFromServer]);
 
   return (
     <main className="section">
-      <h1 className="title">
-        {`${name} (${born} = ${died})`}
-      </h1>
+      {selectedPerson ? (
+        <h1 className="title">{`${selectedPerson.name} (${selectedPerson.born} = ${selectedPerson.died})`}</h1>
```
voronine, line 51: Name (born - died)

voronine (approved):
All ok, one moment

return h1 => Name (born - died),

src/App.tsx
```diff
@@ -1,56 +1,76 @@
...
+  const [query, setQuery] = useState('');
+  const [selectedPerson, setSelectedPerson] = useState<Person | null>(null);
+  const [isDropdownVisible, setIsDropdownVisible] = useState(false);
+  const [appliedQuery, setAppliedQuery] = useState('');
+
+  const applyQuery = useCallback(
+    debounce((value) => {
+      setAppliedQuery(value);
+      setIsDropdownVisible(true);
+    }, 1000),
```
alexander-ignatow, line 19: Per task - debounce time should be the component property
You will need some refactoring to do it (see comment above)

src/App.tsx
alexander-ignatow: Please, fix per @voronine comment

src/App.tsx
alexander-ignatow: Use callback is unnecessary here

alexander-ignatow (changes requested):
Good, but you need to refactor your solution and create dedicated `Autocomplete` component with all your subcomponents and props described in the task and use it in the `App`

PR #232, opened 2023-09-11
added ReactAutocomplete solution

NazarVotkanych:
Demo: https://NazarVotkanych.github.io/react_autocomplete/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,43 +1,12 @@
-import React, { useCallback, useMemo, useState } from 'react';
+import React, { useState } from 'react';
 import './App.scss';
-import debounce from 'lodash.debounce';
-import { peopleFromServer } from './data/people';
-import { DropDownMenu } from './components/DropDownMenu/DropDownMenu';
 import { Person } from './types/Person';
+import { Autocomplete } from './components/Autocomplete/Autocomplete';
 
 export const App: React.FC = () => {
-  const [query, setQuery] = useState('');
   const [selectedPerson, setSelectedPerson] = useState<Person | null>(null);
-  const [isInputFocused, setInputFocused] = useState(false);
-  const [appliedQuery, setAppliedQuery] = useState('');
 
-  const handleInputFocus = () => {
-    setInputFocused(true);
-  };
-
-  const applyQuery = useCallback(
-    debounce(setAppliedQuery, 1000),
-    [],
-  );
-
-  const handleQueryChange = (event: React.ChangeEvent<HTMLInputElement>) => {
-    setQuery(event.target.value);
-    applyQuery(event.target.value);
-  };
-
-  const handlePersonSelect = useCallback((person: Person) => {
-    setSelectedPerson(person);
-    setQuery(person.name);
-    setAppliedQuery(person.name);
-    setInputFocused(false);
-  }, []);
-
-  const filteredPersons = useMemo(() => {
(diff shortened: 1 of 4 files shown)


src/App.tsx
```diff
@@ -1,14 +1,50 @@
...
+
+  const applyQuery = useCallback(
+    debounce(setAppliedQuery, 1000),
+    [],
+  );
+
+  const handleQueryChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+    setQuery(event.target.value);
+    applyQuery(event.target.value);
+  };
```
irahos, line 26: you should hide suggestions on text change according to the task

src/components/DropDownMenu/DropDownMenu.tsx
```diff
@@ -0,0 +1,38 @@
...
+          {people?.length
+            ? (
+              people.map((person) => (
+                <div
+                  className="dropdown-item"
+                  key={person.slug}
+                  onClick={() => onSelect(person)}
+                >
+                  <p className="has-text-link">{person.name}</p>
+                </div>
```
irahos, line 26: you could have created a separate component for this, f.e. DropdownItem

irahos (approved):
Left some comments, please check them😊
other than that great job👏

src/App.tsx
alexander-ignatow: This comment makes sense

src/App.tsx
```diff
@@ -1,14 +1,50 @@
...
+  const [selectedPerson, setSelectedPerson] = useState<Person | null>(null);
+  const [isInputFocused, setInputFocused] = useState(false);
+  const [appliedQuery, setAppliedQuery] = useState('');
+
+  const handleInputFocus = () => {
+    setInputFocused(true);
+  };
+
+  const applyQuery = useCallback(
+    debounce(setAppliedQuery, 1000),
```
alexander-ignatow, line 19: Per task: this should be set up via props

src/App.tsx
```diff
@@ -1,14 +1,50 @@
...
+  };
+
+  const handlePersonSelect = useCallback((person: Person) => {
+    setSelectedPerson(person);
+    setQuery(person.name);
+    setAppliedQuery(person.name);
+    setInputFocused(false);
+  }, []);
+
+  const filteredPersons = useMemo(() => {
```
alexander-ignatow, line 35: This `memo` makes sense, while other memoization most probably not.

alexander-ignatow (changes requested):
Good job!

However, some refactoring is needed
1. You should create a separate component `Autocomplete`, with props as described per task and showcase it with your `App`
2. Don't overuse memoization with `React.memo` and `useCallback` hook. In your case it does not give you benefits - even hurts the app's performance.

src/components/Autocomplete/Autocomplete.tsx
```diff
@@ -0,0 +1,73 @@
...
+  return (
+    <div className="dropdown is-active">
+      <div className="dropdown-trigger">
+        <input
+          type="text"
+          placeholder="Enter a part of the name"
+          className="input"
+          value={query}
+          onChange={handleQueryChange}
+          onFocus={handleInputFocus}
```
alexander-ignatow, line 61: Would be nice to handle `onBlur` and hide the dropdown when input loses the focus

alexander-ignatow (approved):
LGTM

Please - see the comments below to consider them for the future

PR #234, opened 2023-09-11
add task solution

ydratsevska:
[DEMO LINK](https://ydratsevska.github.io/react_autocomplete/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,56 +1,62 @@
-import React from 'react';
+import React, { useState, useCallback } from 'react';
+import { debounce } from 'lodash';
+import classNames from 'classnames';
 import './App.scss';
-import { peopleFromServer } from './data/people';
+import { Person } from './types/Person';
+import { DropdownMenu } from './components/DropdownMenu';
 
 export const App: React.FC = () => {
-  const { name, born, died } = peopleFromServer[0];
+  const [query, setQuery] = useState('');
+  const [selectedPerson, setSelectedPerson] = useState<Person | null>(null);
+  const [isFocused, setIsFocused] = useState(false);
+  const [appliedQuery, setAppliedQuery] = useState('');
+
+  const applyQuery = useCallback(debounce(setAppliedQuery, 1000), []);
+
+  const handelQueryChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+    setQuery(event.target.value);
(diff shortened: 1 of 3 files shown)


polinavafik (approved):
Everything is working as expected!
ona small note

src/App.tsx
```diff
@@ -1,56 +1,62 @@
-import React from 'react';
+import React, { useState, useCallback } from 'react';
+import { debounce } from 'lodash';
```
BudnikOleksii, line 2: You can install and use only debounce instead of all that have "lodash". https://www.npmjs.com/package/lodash.debounce

src/components/DropdownMenu.tsx
```diff
@@ -0,0 +1,59 @@
...
+import { Person } from '../types/Person';
+import { peopleFromServer } from '../data/people';
+
+type Props = {
+  onSelect: (person: Person) => void;
+  appliedQuery: string;
+};
+
+function getFilteredPeople(people: Person[], query: string): Person[] {
+  const filteredPeople = [...people].filter(person => {
```
BudnikOleksii, line 12:
```suggestion
  const filteredPeople = people.filter(person => {
```
Filter method returns a new array, you don't need to copy

src/App.tsx
```diff
@@ -1,56 +1,62 @@
...
+  const applyQuery = useCallback(debounce(setAppliedQuery, 1000), []);
+
+  const handelQueryChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+    setQuery(event.target.value);
+    applyQuery(event.target.value);
+    setIsFocused(true);
+    if (!event.target.value) {
+      setSelectedPerson(null);
+      setIsFocused(false);
+    }
```
BudnikOleksii, line 23:
```suggestion
    const { value } = event.target;
    
    setQuery(value);
    applyQuery(value);
    setIsFocused(true);
    
    if (!event.target.value) {
      setSelectedPerson(null);
      setIsFocused(false);
    }
```
You can use destructing and don't repeat the same twice. Also, add blank lines to group operations for better readability

BudnikOleksii (approved):
Good job👍
Just remember if you can install only part of a library for your needs, use this opportunity.
For this task you need only debounce, so you can install `lodash.debounce` instead of lodash, it reduces your bundle size.

PR #237, opened 2023-09-11
add task solution

dpidlutska:
[DEMO LINK](https://dpidlutska.github.io/react_autocomplete/)


src/DropDown/DropDown.tsx
```diff
@@ -0,0 +1,102 @@
...
+
+    return person.name.trim().toLowerCase().includes(normalizedQuery);
+  })), [apliedQuery, people]);
+
+  const resultToShow = (visiblePeople.length > 0);
+
+  const isPeopleListShown = !selectedPerson
+  && resultToShow
+  && isFocused;
+
```
ivannaromanovych, line 37: I don't understand why do you need so many checks? You can just show no matching message or visiblePeople (depending on the list length) if input is focused
Maybe because of thet it's not working correctly
After i seacrh non-existing person image
It's not showing the list if I erased input 
image

ivannaromanovych (changes requested):
Good job, but it needs few minor fixes to work perfectly

src/DropDown/DropDown.tsx
dpidlutska: Actually filtering works correctly, if enter some invalid input and than delete it: https://gyazo.com/87be1429ea480fe7475372f6820cb403

I add addtitional check to hide list if person is selected. Other checks are nessesary - to check if filtered list has length and if input is focused. I create variables for it to make code more clear.
I don't use conditional operator in this case, because if filtered list is empty, so message 'No matching suggestions' to be shown, it will be shown like list element - on list background. And it doesn't look good, as for me it should be just paragraph under input.

src/DropDown/DropDown.tsx
```diff
@@ -0,0 +1,102 @@
+import React, { useState, useCallback, useMemo } from 'react';
+import debounce from 'lodash.debounce';
+import cn from 'classnames';
+
+import { Person } from '../types/Person';
+
+type Props = {
+  people: Person[];
+  selectedPerson: Person | null;
```
alexander-ignatow, line 9: It is up to you to decide - but if I needed to save the selected person, I would do it inside `DropDown` component

src/DropDown/DropDown.tsx
```diff
@@ -0,0 +1,102 @@
...
+  people,
+  selectedPerson,
+  onSelect = () => {},
+  delay,
+}) => {
+  const [query, setQuery] = useState('');
+  const [apliedQuery, setApliedQuery] = useState('');
+  const [isFocused, setIsFocused] = useState(false);
+
+  const applyQuery = useCallback(debounce(setApliedQuery, delay), []);
```
alexander-ignatow, line 24: This `useCallback` is not needed

src/DropDown/DropDown.tsx
```diff
@@ -0,0 +1,102 @@
...
+      <div className="dropdown-trigger">
+        <input
+          type="text"
+          placeholder="Enter a part of the name"
+          className="input"
+          value={query}
+          onChange={queryHandler}
+          onClick={() => setIsFocused(true)}
+          onBlur={() => setIsFocused(false)}
+          disabled={selectedPerson !== null}
```
alexander-ignatow, line 64: This is a bit strange from the point of the user, so I am not sure of it is a good idea to do so

src/DropDown/DropDown.tsx
```diff
@@ -0,0 +1,102 @@
...
+                sex,
+                name,
+              } = person;
+
+              return (
+                <a
+                  href={`#${slug}`}
+                  className="dropdown-item"
+                  key={slug}
+                  onMouseDown={(event) => selectHander(event, person)}
```
alexander-ignatow, line 83: Please, use `onClick` event here

alexander-ignatow (changes requested):
Good job, but some changes required

Don't block the input once the person is selected - see this [example](https://mui.com/material-ui/react-autocomplete/)

src/DropDown/DropDown.tsx
```diff
@@ -0,0 +1,115 @@
...
+    event: React.MouseEvent<HTMLAnchorElement>,
+    person: Person,
+  ) => {
+    event.preventDefault();
+    onSelect(person);
+    setIsFocused(false);
+    setQuery(person.name);
+  };
+
+  const resetHandler = () => {
```
alexander-ignatow, line 49: this does not clear the search query correct

steps to reproduce:
input 'car' - see dropdown changes
use the reset button in the input - input is cleared
now click into the input to show the dropdown - you see dropdown filtered on "car" with empty input

alexander-ignatow: That happens because you have `selectedPeople` memoized, but their dependencies are not changed in scenario above

src/DropDown/DropDown.tsx
alexander-ignatow: Still `onMouseDown` event - pleas switch to `onClick`

alexander-ignatow (changes requested):
Good progress! let's fix the remaining issues

src/DropDown/DropDown.tsx
dpidlutska: If I use onClick, it doesn't work here. Is this fix critical? I was on QnA with Yuriy and he told it's good way to fix not working onClick with using onMouseDown here.

src/DropDown/DropDown.tsx
```diff
@@ -0,0 +1,116 @@
...
+  const [apliedQuery, setApliedQuery] = useState('');
+  const [isFocused, setIsFocused] = useState(false);
+
+  const applyQuery = debounce(setApliedQuery, delay);
+
+  const visiblePeople = useMemo(() => (people.filter(person => {
+    const normalizedQuery = apliedQuery.trim().toLowerCase();
+
+    return person.name.trim().toLowerCase().includes(normalizedQuery);
+  })), [apliedQuery, people]);
```
BudnikOleksii, line 28:
```suggestion
  const visiblePeople = useMemo(
    () => (people.filter(person => {
      const normalizedQuery = apliedQuery.trim().toLowerCase();

      return person.name.trim().toLowerCase().includes(normalizedQuery);
    })),
    [apliedQuery, people],
  );
```

BudnikOleksii (approved):
Good work 👍

PR #238, opened 2023-09-11
Solution🎀

polinavafik:
[DEMO LINK](https://polinavafik.github.io/react_autocomplete/)


vserhiichuk (approved):
It looks good, but I just didn't understand why you use the `delay`  in props.
And also, when you have already chosen a person and change him (enter a new one), everything should also work

src/components/Dropdown.tsx
```diff
@@ -0,0 +1,123 @@
...
+} from 'react';
+import debounce from 'lodash.debounce';
+import classNames from 'classnames';
+import { Person } from '../types/Person';
+
+interface Props {
+  people: Person[];
+  delay: number;
+  onSelected?: (user: Person) => void,
+}
```
alexander-ignatow, line 16: 👍

src/components/Dropdown.tsx
```diff
@@ -0,0 +1,123 @@
...
+
+export const Dropdown: React.FC<Props> = React.memo(({
```
alexander-ignatow, line 18: I would say `Autocomplete` or `AutocompletePeople` would be a better name

alexander-ignatow, line 18: `React.memo` is a bad idea for such component

Let the end user decide if it is needed

src/components/Dropdown.tsx
```diff
@@ -0,0 +1,123 @@
...
+    debounce(setAppliedQuery, delay),
+    [],
+  );
+
+  const handleQueryChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+    setQuery(event.target.value.trimStart());
+    applyQuery(event.target.value);
+    if (query) {
+      setIsProcessing(true);
+    }
```
alexander-ignatow, line 38: You got the idea of debounce wrong

You should not emulate the call to the server - rather it limits the number of calls to the underlying API

So you don't need  `isProcessing` state at all

Most autocomplete components just show the current list, while the person is typing and update the value when the user stops typing (i.e. debounce the call)

src/components/Dropdown.tsx
```diff
@@ -0,0 +1,123 @@
...
+                    {person.name}
+                  </p>
+                </a>
+              ))
+            )
+            : (
+              <div className="p-2 notification is-info has-text-ligth">
+                Processing
+              </div>
+            )}
```
alexander-ignatow, line 112: As mentioned above - don't do this, just show the current value of the filter

alexander-ignatow (changes requested):
Good job, but some updates are needed

And one bug was spotted - the dropdown is opened once the page is loaded (even before input gets the focus)

polinavafik:
@alexander-ignatow 

> And one bug was spotted - the dropdown is opened once the page is loaded (even before input gets the focus)
its actually not a bug, I did that on purpose with useRef and useEffect, but I can remove it

src/components/Dropdown.tsx
polinavafik: since we return a <div className='dropdown'>, I named the component this way + its easy to understand what kind of stuff we get on our page

For me, it would be hard to understand what kind of component "insides" we are getting, with a name "Autocomplete"

So I would prefer to leave the name as it is

src/components/Dropdown.tsx
```diff
@@ -0,0 +1,109 @@
...
+            <a
+              className="dropdown-item"
+              href="/"
+              key={person.slug}
+              onMouseDown={(event) => handleMouseDown(event, person)}
+            >
+              <p
+                className={classNames({
+                  'has-text-link': person.sex === 'm',
+                  'has-text-danger': person.sex === 'f',
```
roman-mirzoian, line 92: Avoid using "magic values".

src/components/Dropdown.tsx
```diff
@@ -0,0 +1,109 @@
...
+import classNames from 'classnames';
+import { Person } from '../types/Person';
+
+interface Props {
+  people: Person[];
+  delay: number;
+  onSelected?: (user: Person) => void,
+}
+
+export const Dropdown: React.FC<Props> = ({
```
roman-mirzoian, line 18: https://github.com/mate-academy/react_autocomplete/pull/238#discussion_r1324816582
Dropdown is a common name for a component that has a dropdown list.

Therefore, I agree with Alexander that it is worth giving a more personalized name or combining both options: `DropdownAutocomplete`, for example.
image

Also, you can see the approach of the developers of the MUI library, that they would also call such a component `Autocomplete` :)

image

roman-mirzoian (changes requested):
Now the message about the mismatch can only be seen with the selected person if you click outside the input and then click again on it:
https://gyazo.com/69e889b298b07baada7a6845d1250f2b

src/components/Dropdown.tsx
polinavafik: :( ok

src/components/Autocomplete.tsx
```diff
@@ -0,0 +1,106 @@
...
+  useCallback,
+  useMemo,
+  useState,
+} from 'react';
+import debounce from 'lodash.debounce';
+import classNames from 'classnames';
+import { Person } from '../types/Person';
+
+const MALE = 'm';
+const FEMALE = 'f';
```
roman-mirzoian, line 11: If we use TypeScript it can be specified as Enum.

src/components/Autocomplete.tsx
```diff
@@ -0,0 +1,106 @@
...
+    [],
+  );
+
+  const handleQueryChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+    setQuery(event.target.value.trimStart());
+    applyQuery(event.target.value);
+    setHasFocus(true);
+    if (!event.target.value) {
+      setHasFocus(false);
+    }
```
roman-mirzoian, line 39: You use `event.target.value` three times so it can be put in a separate variable.

src/components/Autocomplete.tsx
```diff
@@ -0,0 +1,106 @@
...
+  };
+
+  const filteredPeople = useMemo(() => {
+    return people.filter(
+      // eslint-disable-next-line max-len
+      person => person.name.toLowerCase().includes(appliedQuery.trim().toLowerCase()),
+    );
```
roman-mirzoian, line 46: It is bad practice to disable the linter, so you should try to write the code so that everything works:
```suggestion
    return people.filter(({ name }) => {
        const preparedName = name.toLowerCase();
        const preparedQuery = appliedQuery.trim().toLowerCase();
        return preparedName.includes(preparedQuery);
      }
    );
```

src/components/Autocomplete.tsx
```diff
@@ -0,0 +1,106 @@
...
+          value={query}
+          onChange={handleQueryChange}
+          onFocus={() => setHasFocus(true)}
+          onBlur={() => setHasFocus(false)}
+        />
+      </div>
+
+      <div className="dropdown-menu" role="menu">
+        <div className="dropdown-content has-text-link">
+          {filteredPeople.length !== 0
```
roman-mirzoian, line 78:
```suggestion
          {filteredPeople.length
```

src/App.tsx
```diff
@@ -1,57 +1,28 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
 import { peopleFromServer } from './data/people';
+import { Autocomplete } from './components/Autocomplete';
+import { Person } from './types/Person';
+
+const DELAY = 500;
```
roman-mirzoian, line 7:
```suggestion
const SEARCH_DELAY = 500;
```

roman-mirzoian (changes requested):
Almost done!

src/components/Autocomplete.tsx
```diff
@@ -0,0 +1,112 @@
...
+  const [appliedQuery, setAppliedQuery] = useState('');
+  const [hasFocus, setHasFocus] = useState(false);
+
+  const applyQuery = useCallback(
+    debounce(setAppliedQuery, delay),
+    [],
+  );
+
+  const handleQueryChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+    const inputValue = event.target.value;
```
roman-mirzoian, line 36:
```suggestion
    const query = event.target.value;
```

roman-mirzoian (approved):
Well done 👍

PR #241, opened 2023-09-11
implementing autocomplete

Yevheniia-Sid:
[DEMO LINK](https://Yevheniia-Sid.github.io/react_autocomplete/)


src/App.tsx
```diff
@@ -1,57 +1,31 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
 import { peopleFromServer } from './data/people';
+import { Person } from './types/Person';
+import { Dropdown } from './components/Dropdown';
+
+const DELAY_MS = 1000;
```
roman-mirzoian, line 7:
```suggestion
const SEARCH_DELAY = 1000;
```

src/App.tsx
```diff
@@ -1,57 +1,31 @@
...
-              <p className="has-text-danger">Petronella de Decker</p>
-            </div>
+      {selectedPerson
+        ? (
+          <h1 className="title">
+            {`${selectedPerson.name} (${selectedPerson.born} = ${selectedPerson.died})`}
+          </h1>
+        ) : (
+          <h1>No selected person</h1>
+        )}
```
roman-mirzoian, line 21: It is not necessary to write the same tag twice, it is enough to write it once, move the condition for the text to the middle of the tag, and define the className through the classnames library.

src/components/Dropdown.tsx
```diff
@@ -0,0 +1,99 @@
+import { useCallback, useState } from 'react';
+import classNames from 'classnames';
+import debounce from 'lodash.debounce';
+import { Person } from '../types/Person';
+
+type Props = {
+  people: Person[],
+  delay: number,
```
roman-mirzoian, line 8:
```suggestion
  searchdDelay: number,
```

src/components/Dropdown.tsx
```diff
@@ -0,0 +1,99 @@
...
+                    className="dropdown-item"
+                    key={person.slug}
+                    role="button"
+                    onClick={() => handlePersonClick(person)}
+                    onKeyDown={() => { }}
+                    tabIndex={0}
+                  >
+                    <p className={classNames(
+                      { 'has-text-link': person.sex === 'm' },
+                      { 'has-text-danger': person.sex === 'f' },
```
roman-mirzoian, line 79: Avoid using magic values.

src/components/Dropdown.tsx
```diff
@@ -0,0 +1,99 @@
...
+          <div className="dropdown-content">
+            {personList.length
+              ? (
+                personList.map((person) => (
+                  <div
+                    className="dropdown-item"
+                    key={person.slug}
+                    role="button"
+                    onClick={() => handlePersonClick(person)}
+                    onKeyDown={() => { }}
```
roman-mirzoian, line 74: Why is there a handler with an empty callback?

src/components/Dropdown.tsx
```diff
@@ -0,0 +1,99 @@
...
+                    tabIndex={0}
```
roman-mirzoian, line 75: Do you really need this attribute now?

roman-mirzoian (changes requested):

src/components/Dropdown.tsx
Yevheniia-Sid: > Do you really need this attribute now?

same as previous

src/components/Dropdown.tsx
```diff
@@ -0,0 +1,106 @@
...
+  const applyQuery = useCallback(
+    debounce(setAppliedQuery, searchDelay),
+    [],
+  );
+
+  const handelQueryChange = (
+    event: React.ChangeEvent<HTMLInputElement>,
+  ) => {
+    setQuery(event.target.value);
+    applyQuery(event.target.value);
```
BudnikOleksii, line 54: You can get value with destructing and in this case, don't repeat `event.target.value` two times

src/App.tsx
```diff
@@ -1,57 +1,32 @@
...
+  const [selectedPerson, setSelectedPerson] = useState<Person | null>(null);
 
   return (
     <main className="section">
-      <h1 className="title">
-        {`${name} (${born} = ${died})`}
+      <h1 className={classNames({ title: selectedPerson })}>
+        {selectedPerson
+          ? (
+            `${selectedPerson.name} (${selectedPerson.born} = ${selectedPerson.died})`
```
BudnikOleksii, line 18:
```suggestion
            `${selectedPerson.name} (${selectedPerson.born} - ${selectedPerson.died})`
```
I think it should be `-` instead of `=`
image

BudnikOleksii (approved):
Good work 👍

PR #248, opened 2023-09-14
add solution

purpleFade:
[DEMO LINK](https://purplefade.github.io/react_autocomplete/)


sTorba24 (approved):
Well done generally!

you can improve it

when smth selected it should show this option on focus, or should be visible all options and selected marked somehow

image

PR #249, opened 2023-09-14
task solution was added

Liliya-Kalinichenko:
[DEMO LINK](https://Liliya-Kalinichenko.github.io/react_autocomplete/)

### DIFF
--- a/src/Autocomplete.tsx
+++ b/src/Autocomplete.tsx
@@ -19,7 +19,10 @@ export const Autocomplete: React.FC<Props> = ({
   const [query, setQuery] = useState('');
   const [appliedQuery, setAppliedQuery] = useState('');
 
-  const applyQuery = useCallback(debounce(setAppliedQuery, delay), [delay]);
+  const applyQuery = useCallback(debounce((value: string) => {
+    setAppliedQuery(value);
+    setIsDropdownOpen(true);
+  }, delay), [delay]);
 
   const filteredPeople = useMemo(() => {
     const normalizedQuery = appliedQuery.toLowerCase().trim();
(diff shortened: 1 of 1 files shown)


src/Autocomplete.tsx
```diff
@@ -0,0 +1,83 @@
...
+  };
+
+  const handleInputChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+    setQuery(event.target.value);
+    applyQuery(event.target.value);
+    onSelected(null);
+  };
+
+  return (
+    <>
```
TarasHoliuk, line 45: it seems like [fragment](https://react.dev/reference/react/Fragment) is redundant

src/Autocomplete.tsx
```diff
@@ -0,0 +1,83 @@
...
+                  href="/"
+                  key={person.slug}
+                  onMouseDown={() => handleSelect(person)}
+                >
+                  <p className="has-text-link">{person.name}</p>
+                </a>
+              ))}
+            </div>
+          ) : (
+            <p className="help is-danger">No matching suggestions</p>
```
TarasHoliuk, line 77: I'd show it inside the popup - and not as an error, like one of suggestions, with black text

but it's not critical, and there are no such requirements in the task, so up to you

src/Autocomplete.tsx
```diff
@@ -0,0 +1,83 @@
...
+    );
+  }, [appliedQuery, people]);
+
+  const handleSelect = (person: Person) => {
+    setQuery(person.name);
+    onSelected(person);
+    setIsDropdownOpen(false);
+  };
+
+  const handleInputChange = (event: React.ChangeEvent<HTMLInputElement>) => {
```
TarasHoliuk, line 38: let's implement this part:
image

feel free to ask if you need help :slightly_smiling_face:

TarasHoliuk (changes requested):

TarasHoliuk (approved):
Fixed so fast :fire:

PR #254, opened 2023-09-18
Add task solution

klaudia-klis:
[DEMO LINK](https://klaudia-klis.github.io/react_autocomplete/)

### DIFF
--- a/src/components/Dropdown.tsx
+++ b/src/components/Dropdown.tsx
@@ -1,0 +1,74 @@
+import { useState } from 'react';
+import classNames from 'classnames';
+import { useDebounce } from 'use-debounce';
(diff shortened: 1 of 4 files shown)


src/components/Dropdown.tsx
```diff
@@ -0,0 +1,84 @@
+import { useState } from 'react';
+import classNames from 'classnames';
+import { useDebounce } from 'use-debounce';
```
choeqq, line 3: I think creating debounce hook yourself would be pretty useful, instead of installing a whole dependency

choeqq (changes requested):

choeqq (approved):
LGTM 🚀

PR #256, opened 2023-09-18
Add task solution

NikitaOrliak:
[DEMO LINK](https://NikitaOrliak.github.io/react_autocomplete/)

Honestly, I don't understand 3 things:

1.  If I need to reset h1 to 'No selected person' with an empty title, where should I check this and throw null to the App.tsx? I added it in 25-27 lines, but I think it's incorrect.
2. Do I need to use 'useRef' in the input? And why? And, maybe, I should use memo to all my Autocomplete func?
3. Why do I need to create another appliedQueary state for debounce? Why can't I use debounce with my setInputText? It's just doesn't work and I don't understand why, but with additional state it works correctly.

Thanks!


src/components/Autocomplete.tsx
```diff
@@ -0,0 +1,105 @@
...
+          className="input"
+          onFocus={openMenu}
+          onBlur={() => setTimeout(closeMenu, 200)}
+        />
+      </div>
+
+      <div className="dropdown-menu" role="menu">
+        <div className="dropdown-content">
+          {
+            sortedListOfPeople.length > 0
```
volodymyr-soltys97, line 83:
```suggestion
            sortedListOfPeople.length
```

src/components/Autocomplete.tsx
```diff
@@ -0,0 +1,105 @@
...
+              ? (
+                sortedListOfPeople.map((person) => (
+                  <a
```
volodymyr-soltys97, line 86: You can use a destructuring for `person`

src/App.tsx
```diff
@@ -1,57 +1,34 @@
...
+
+  const getYear = useCallback((person: Person) => {
+    const { name, born, died } = person;
+
+    return `${name} (${born} - ${died})`;
+  }, [selectedPerson]);
+
+  const title = selectedPerson !== null
+    ? getYear(selectedPerson)
+    : 'No selected person';
```
volodymyr-soltys97, line 18:
```suggestion
  const title = selectedPerson
    ? getYear(selectedPerson)
    : 'No selected person';
```

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code
1. It's clearly not a good idea to check for null
2. You can use useMemo for your function, it will be better

lerastarynets (approved):
Great job overall! Left one tiny suggestion for the improvement

PR #259, opened 2023-09-19
add task solution

denlubn:
[DEMO LINK](https://denlubn.github.io/react_autocomplete/)


volodymyr-soltys97 (changes requested):
Great work 👍 
Let's improve
1. Add `cursor: pointer` for items
image

2. The list should only be shown when the input is in focus
image

PR #260, opened 2023-09-20
Develop

reseneweb:
https://reseneweb.github.io/react_autocomplete/


sTorba24 (changes requested):
make search not case sensitive 
image

Also hide options on select

image

PR #273, opened 2023-09-30
Solution

Reject-4444:
https://reject-4444.github.io/react_autocomplete/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,14 +1,42 @@
-import React from 'react';
+import React, { useCallback, useMemo, useState } from 'react';
 import './App.scss';
+import debounce from 'lodash.debounce';
 import { peopleFromServer } from './data/people';
+import { DropDownItem } from './components/DropDownItem';
+import { Person } from './types/Person';
 
 export const App: React.FC = () => {
-  const { name, born, died } = peopleFromServer[0];
+  const [query, setQuery] = useState('');
+  const [appliedQuery, setAppliedQuery] = useState('');
+  const [selectedPerson, setSelectedPerson] = useState<Person | null>(null);
+
+  const delay = 1000;
+
+  const applyQuery = useCallback(debounce(setAppliedQuery, delay), [delay]);
+
+  const handleQueryChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+    setQuery(event.target.value);
+    applyQuery(event.target.value);
+  };
+
+  const handlePersonSelect = (person: Person) => {
+    setSelectedPerson(person);
+    setQuery('');
+  };
+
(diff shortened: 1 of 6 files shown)


src/App.tsx
```diff
@@ -1,14 +1,42 @@
...
+        .toLowerCase().includes(appliedQuery.toLowerCase()),
+    );
+  }, [appliedQuery]);
 
   return (
     <main className="section">
       <h1 className="title">
-        {`${name} (${born} = ${died})`}
+        {selectedPerson
+          ? `${selectedPerson.name} (${selectedPerson.born} - ${selectedPerson.died})`
```
loralevitska, line 38: DRY - don't repeat yourself. Destructure `selectedPerson`

loralevitska (changes requested):
Keep up the good work! Some comments to fix:
- the list should be hidden by default and be visible by after activate search field 
image
image
- close list after selected person
image
- use [string.trim()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/trim) for search field 
image

Reject-4444:
Please look at this code and help me to fix a bugs. Problem is that when i onBlur input and change state isFocused(false) i don't have possibility to chose any person because list had disappeared before i clicked on button

etojeDenys:
> https://reject-4444.github.io/react_autocomplete/

please, feel free to ask about that in chat

etojeDenys (changes requested):
i should have the ability to close the list just by unfocusing the button

Reject-4444:
i've already done all the tasks..i hope so..)

src/App.tsx
```diff
@@ -1,54 +1,97 @@
...
+    if (dropDownRef.current && !dropDownRef.current
+      .contains(event.target as Node)) {
+      setIsFocused(false);
+    } else {
+      setIsFocused(true);
+    }
+  };
+
+  useEffect(() => {
+    document.addEventListener('mousedown', handleClick);
```
anastasiiavorobiova, line 66: Consider using React features instead of standard DOM event handlers

anastasiiavorobiova (changes requested):
Great work! 
Screenshot 2023-10-01 at 17 31 41
Screenshot 2023-10-01 at 17 33 55

Consider hiding an empty list on blur
Screenshot 2023-10-01 at 17 31 49
Add paddings for the message
Screenshot 2023-10-01 at 17 32 12
The person's name should be clickable
Screenshot 2023-10-01 at 17 32 40
If the person is selected the one matching suggestion should be shown

anastasiiavorobiova (changes requested):
The previous comments were not fixed

xiimiik (approved):
nice work!

PR #275, opened 2023-10-01
Develop

aniashev:
[DEMO LINK](https://aniashev.github.io/react_autocomplete/)


VitaliyBondarenko1982 (changes requested):
Dropdown should be not visible when input is not in focus.
Add placeholder for input.
chrome-capture-2023-9-1 (1)

src/App.tsx
```diff
@@ -1,55 +1,99 @@
...
     <main className="section">
       <h1 className="title">
-        {`${name} (${born} = ${died})`}
+        { selectedPerson
+          ? `${selectedPerson.name} (${selectedPerson.born} - ${selectedPerson.died})`
+          : 'No selected person'}
       </h1>
 
-      <div className="dropdown is-active">
+      <div className={getDropdownClass(dropdownFocused)}>
```
etojeDenys, line 84: you can just use classNames package for such cases instead of the funciton

src/index.tsx
```diff
@@ -2,4 +2,4 @@ import ReactDOM from 'react-dom';
 import 'bulma/css/bulma.css';
 import { App } from './App';
 
-ReactDOM.render(<App />, document.getElementById('root'));
+ReactDOM.render(<App delay={500} />, document.getElementById('root'));
```
etojeDenys, line 5: let's increase the default delay, and better to remove delay as a prop and just add it as a constant in app.tsx

etojeDenys (changes requested):

src/App.tsx
```diff
@@ -1,55 +1,93 @@
...
+    }
+
+    setDropdownFocused(true);
+  };
+
+  const handleBlur = () => {
+    setTimeout(() => {
+      setDropdownFocused(false);
+    }, 200);
+  };
```
VitaliyBondarenko1982, line 62:
```suggestion
  const handleBlur = () => {
      setDropdownFocused(false);
  };
```
we can avoid timeout when will use in PeopleInfo `onMouseDown` event instead `onClick` event.

VitaliyBondarenko1982 (changes requested):
Almost done.
- Suggest to not use timeout (check comment)
- placeholder for input still not added
- check your deploy - dropdown still not hidden when input lost focus.

src/App.tsx
aniashev: I apologize, there was indeed an error with deploy. Thank you for pointing it out!

aniashev:
[DEMO LINK](https://aniashev.github.io/react_autocomplete/)

etojeDenys (approved):
good job

PR #285, opened 2023-10-08
autocomplete

stunnerboyone:
[DEMO LINK](https://stunnerboyone.github.io/react_autocomplete/)


anastasiiavorobiova (changes requested):
GTreat work! 
Screenshot 2023-10-08 at 17 53 01
Consider showing just one match in this case

PR #289, opened 2023-10-11
miss semicolon

Absolution26:
[DEMO LINK](https://Absolution26.github.io/react_autocomplete/)


volodymyr-soltys97 (changes requested):
Great job 🔥 
Let's improve your page
Consider will be better when a person is deleted, the text at the top should also be cleaned
image

volodymyr-soltys97 (approved):
Well done 🔥

PR #292, opened 2023-10-12
add task solution

Anmonnn:
[DEMO LINK](https://Anmonnn.github.io/react_autocomplete/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,14 +1,51 @@
-import React from 'react';
+import React, { useMemo, useRef, useState } from 'react';
 import './App.scss';
 import { peopleFromServer } from './data/people';
+import { Person } from './types/Person';
 
 export const App: React.FC = () => {
-  const { name, born, died } = peopleFromServer[0];
+  const [isInput, setIsInput] = useState('');
+  const [isInputFocused, setInputFocused] = useState(false);
+  const [isAppliedInput, setIsAppliedInput] = useState('');
+  const [isSelectedPerson, setIsSelectedPerson] = useState(peopleFromServer[0]);
+  const [isSelected, setIsSelected] = useState(false);
+  const { name, born, died } = isSelectedPerson;
+
+  const timerId = useRef(0);
+
+  const filteredPeople = useMemo(() => {
+    return peopleFromServer.filter(person => person
+      .name.toLowerCase().includes(isAppliedInput.toLowerCase()));
+  }, [isAppliedInput]);
+
+  const handleInputChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+    setIsInput(event.target.value);
+
+    if (isAppliedInput !== event.target.value) {
+      window.clearTimeout(timerId.current);
+
+      timerId.current = window.setTimeout(() => {
+        setIsAppliedInput(event.target.value);
+      }, 500);
+    }
+  };
+
+  const handleisSelected = (person: Person) => {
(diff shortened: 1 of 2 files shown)


etojeDenys (changes requested):
1. let's hide this block in such cases
image

2. you need to save the person's name in the input value
image

src/App.tsx
```diff
@@ -1,14 +1,52 @@
-import React from 'react';
+import React, { useMemo, useRef, useState } from 'react';
 import './App.scss';
 import { peopleFromServer } from './data/people';
+import { Person } from './types/Person';
 
 export const App: React.FC = () => {
-  const { name, born, died } = peopleFromServer[0];
+  const [isInput, setIsInput] = useState('');
```
etojeDenys, line 7: when you name state or var starting with `is` we expect that it will be the boolean var instead of string, so let's rename it

src/App.tsx
```diff
@@ -1,14 +1,52 @@
...
+  const [isInputFocused, setInputFocused] = useState(false);
+  const [isAppliedInput, setIsAppliedInput] = useState('');
+  const [isSelectedPerson, setIsSelectedPerson] = useState(peopleFromServer[0]);
```
etojeDenys, line 10: same here

src/App.tsx
```diff
@@ -1,14 +1,52 @@
...
+
+  const handleInputChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+    setIsInput(event.target.value);
+
+    if (isAppliedInput !== event.target.value) {
+      window.clearTimeout(timerId.current);
+
+      timerId.current = window.setTimeout(() => {
+        setIsAppliedInput(event.target.value);
+      }, 500);
```
etojeDenys, line 29: let's move this delay logic to some separate function (debounce)

etojeDenys (changes requested):

src/App.tsx
Anmonnn: Why do this if it is only used once?

src/App.tsx
```diff
@@ -1,14 +1,56 @@
...
+  const { name, born, died } = selectedPerson;
+
+  const timerId = useRef(0);
+
+  const filteredPeople = useMemo(() => {
+    return peopleFromServer.filter(person => person
+      .name.toLowerCase().includes(appliedInput.toLowerCase()));
+  }, [appliedInput]);
+
+  function debounce(event: React.ChangeEvent<HTMLInputElement>) {
```
lerastarynets, line 21: it's bad practice to use different approaches among the project, if u use arrow functions everywhere, make ur debounce an arrow func as well

src/App.tsx
```diff
@@ -17,41 +59,42 @@ export const App: React.FC = () => {
...
-            <div className="dropdown-item">
-              <p className="has-text-danger">Petronella de Decker</p>
-            </div>
-
-            <div className="dropdown-item">
-              <p className="has-text-danger">Elisabeth Hercke</p>
+        {isInputFocused && (
+          <div className="dropdown-menu" role="menu">
+            <div className="dropdown-content">
+              {filteredPeople.length > 0 ? (
```
lerastarynets, line 72:
```suggestion
              {!!filteredPeople.length ? (
```

lerastarynets (approved):
Great job overall! Left a few comments to make your code more advanced)

PR #301, opened 2023-10-19
React_Autocomplete

levchenko-dmytro:
[DEMO LINK](https://levchenko-dmytro.github.io/react_autocomplete/)

### DIFF
--- a/src/components/Dropdown/Dropdown.tsx
+++ b/src/components/Dropdown/Dropdown.tsx
@@ -1,0 +1,92 @@
+import cn from 'classnames';
+import './Dropdown.scss';
+import { Person } from '../../types/Person';
+import { useState, useMemo, useCallback } from 'react';
+import { debounce } from '../../services/debounce';
+
+type Props = {
+  peopleList: Person[];
+  onSelected?: (person: Person | null) => void;
+  delay: number;
+};
+
+export const Dropdown: React.FC<Props> = ({
+  peopleList,
+  onSelected = () => {},
+  delay,
+}) => {
+  const [query, setQuery] = useState('');
+  const [visible, setVisible] = useState(false);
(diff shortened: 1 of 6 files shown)


src/components/Dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,90 @@
...
+  delay: number;
+};
+
+export const Dropdown: React.FC<Props> = ({
+  peopleList,
+  onSelected = () => {},
+  delay,
+}) => {
+  const [query, setQuery] = useState('');
+  const [visible, setVisible] = useState(false);
```
loralevitska, line 19: Rename all boolean variables according [tips](https://dev.to/michi/tips-on-naming-boolean-variables-cleaner-code-35ig)

```suggestion
  const [isVisible, setIsVisible] = useState(false);
```

src/components/Dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,90 @@
...
+                    className="dropdown-item dropdown-item--hover"
+                    key={person.slug}
+                    onMouseDown={() => handleSelect(person)}
+                    onKeyDown={() => handleSelect(person)}
+                  >
+                    <p className={cn('has-text-link', {
+                      'has-text-danger': person.sex === 'f',
+                    })}
+                    >
+                      {person.name}
```
loralevitska, line 76: Destructure `person`

loralevitska (changes requested):
A little more and it will be completed

- use [string.trim()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/trim) for search
image

etojeDenys (approved):
looks awesome

PR #306, opened 2023-10-22
add task solution

Sergej161995:
[DEMO LINK](https://Sergej161995.github.io/react_autocomplete/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,14 +1,63 @@
-import React from 'react';
+import React, {
+  useCallback, useMemo, useRef, useState,
+} from 'react';
+
 import './App.scss';
 import { peopleFromServer } from './data/people';
+import { People } from './components/People';
 
-export const App: React.FC = () => {
-  const { name, born, died } = peopleFromServer[0];
+type Props = {
+  delay: number
+};
+
+export const App: React.FC<Props> = ({ delay }) => {
+  const [user, setUser] = useState('');
+  const [appliedQuery, setAppliedQuery] = useState('');
+  const [isFocused, setIsFocused] = useState(false);
+  const [selectedUser, setSelectedUser] = useState('');
+
+  const timerId = useRef(0);
+
+  const findUser = (username: string) => {
+    return peopleFromServer.find(person => person.name === username);
+  };
+
+  const handleQueryChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+    setUser(event.target.value);
+    window.clearTimeout(timerId.current);
+
+    if (user.length <= 1) {
+      setSelectedUser('');
+    }
+
+    timerId.current = window.setTimeout(() => {
+      setAppliedQuery(event.target.value);
+    }, delay);
+  };
+
+  const filteredUsers = useMemo(() => {
+    if (!appliedQuery) {
+      return peopleFromServer;
+    }
+
+    const users = peopleFromServer
+      .filter(person => person.name.toLowerCase()
+        .includes(appliedQuery.toLowerCase()));
+
+    return users;
+  }, [appliedQuery]);
+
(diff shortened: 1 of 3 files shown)


src/App.tsx
```diff
@@ -1,14 +1,63 @@
...
   return (
     <main className="section">
       <h1 className="title">
-        {`${name} (${born} = ${died})`}
+        {selectedUser
+          ? `${findUser(selectedUser)?.name} (${findUser(selectedUser)?.born} = ${findUser(selectedUser)?.died})`
+          : 'No selected person'}
       </h1>
 
       <div className="dropdown is-active">
```
TarasHoliuk, line 63: let's move the entire dropdown to a separate component, and make `delay` its prop instead of the App component

you can move it to `People` component, but rename it to `Dropdown` or `Autocomplete` etc

src/App.tsx
```diff
@@ -1,14 +1,63 @@
...
 import './App.scss';
 import { peopleFromServer } from './data/people';
+import { People } from './components/People';
+
+type Props = {
+  delay: number
+};
+
+export const App: React.FC<Props> = ({ delay }) => {
+  const [user, setUser] = useState('');
```
TarasHoliuk, line 14:
```suggestion
  const [query, setQuery] = useState('');
```
it'll be more accurate

src/App.tsx
```diff
@@ -1,14 +1,63 @@
...
+
+  const findUser = (username: string) => {
+    return peopleFromServer.find(person => person.name === username);
+  };
+
+  const handleQueryChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+    setUser(event.target.value);
+    window.clearTimeout(timerId.current);
+
+    if (user.length <= 1) {
```
TarasHoliuk, line 29:
```suggestion
    if (query) {
```

src/App.tsx
```diff
@@ -17,40 +66,18 @@ export const App: React.FC = () => {
             type="text"
             placeholder="Enter a part of the name"
             className="input"
+            value={user}
+            onChange={event => handleQueryChange(event)}
+            onFocus={() => setIsFocused(true)}
+            onBlur={() => setTimeout(() => setIsFocused(false), 500)}
```
TarasHoliuk, line 72: do you really need a delay here? I'd remove it. Otherwise move the function to a separate variable

src/App.tsx
```diff
@@ -17,40 +66,18 @@ export const App: React.FC = () => {
             type="text"
             placeholder="Enter a part of the name"
             className="input"
+            value={user}
+            onChange={event => handleQueryChange(event)}
```
TarasHoliuk, line 70:
```suggestion
            onChange={handleQueryChange}
```

src/App.tsx
```diff
@@ -1,14 +1,63 @@
...
+  const [isFocused, setIsFocused] = useState(false);
+  const [selectedUser, setSelectedUser] = useState('');
+
+  const timerId = useRef(0);
+
+  const findUser = (username: string) => {
+    return peopleFromServer.find(person => person.name === username);
+  };
+
+  const handleQueryChange = (event: React.ChangeEvent<HTMLInputElement>) => {
```
TarasHoliuk, line 25: let's implement this requirement:
image

src/App.tsx
```diff
@@ -1,14 +1,63 @@
...
+    setUser(event.target.value);
+    window.clearTimeout(timerId.current);
+
+    if (user.length <= 1) {
+      setSelectedUser('');
+    }
+
+    timerId.current = window.setTimeout(() => {
```
TarasHoliuk, line 33: move all "delay" set logic to a separate function with `debounce` name. You can google more about original debounce function

TarasHoliuk (changes requested):

PR #311, opened 2023-10-24
solution

vshepel:
[DEMO LINK](https://vshepel.github.io/react_autocomplete/)


xiimiik (changes requested):
when i select someone i have redirect to this page, it should be like that
image

etojeDenys (approved):
awesome loader

PR #317, opened 2023-10-24
solution

vlevik:
[DEMO LINK](https://vlevik.github.io/react_autocomplete/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,14 +1,46 @@
-import React from 'react';
+import React, { useState, useMemo, useCallback } from 'react';
 import './App.scss';
 import { peopleFromServer } from './data/people';
+import { Person } from './types/Person';
+import { PeopleList } from './components/PeopleList/PeopleList';
+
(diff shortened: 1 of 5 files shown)


src/App.tsx
```diff
@@ -1,56 +1,72 @@
...
-            </div>
-
-            <div className="dropdown-item">
-              <p className="has-text-danger">Petronella de Decker</p>
-            </div>
-
-            <div className="dropdown-item">
-              <p className="has-text-danger">Elisabeth Hercke</p>
-            </div>
+        {shownList && (
```
YaroslavYarynych, line 60: You need to add "No matching suggestions" if we don`t find any person.
image

Like this: 
image

YaroslavYarynych (approved):
Good job! Let`s improve your code a little bit and it will be great!

polosanya (approved):
Good job!
It would be great to make the whole row of the list clickable for better UX
image

PR #318, opened 2023-10-24
`Autocomplete implemented`

sviatikkondur:
Hello, check my work!
[DEMO LINK](https://sviatikkondur.github.io/react_autocomplete/)

### DIFF
--- a/src/components/Dropdown.tsx
+++ b/src/components/Dropdown.tsx
@@ -1,3 +1,4 @@
+/* eslint-disable react-hooks/exhaustive-deps */
 /* eslint-disable jsx-a11y/click-events-have-key-events */
 /* eslint-disable jsx-a11y/no-static-element-interactions */
 /* eslint-disable jsx-a11y/anchor-is-valid */
--- a/src/components/Dropdown.tsx
+++ b/src/components/Dropdown.tsx
@@ -1,0 +1,108 @@
+/* eslint-disable jsx-a11y/click-events-have-key-events */
+/* eslint-disable jsx-a11y/no-static-element-interactions */
+/* eslint-disable jsx-a11y/anchor-is-valid */
+import React, { useState, useCallback } from 'react';
+import cn from 'classnames';
+import debounce from 'lodash.debounce';
+import { Person } from '../types/Person';
+
+type Props = {
+  peopleFromServer: Person[];
+  onSelected: (person: Person) => void;
+};
+
+export const Dropdown: React.FC<Props> = ({
+  peopleFromServer,
+  onSelected = () => {},
+}) => {
+  const [isActive, setIsActive] = useState(false);
+  const [query, setQuery] = useState('');
+  const [appliedQuery, setAppliedQuery] = useState('');
+
+  const applyQuery = useCallback(
+    debounce((value) => {
+      setAppliedQuery(value.trim() !== '' ? value : '');
+      setIsActive(true);
+    }, 1000),
+    [setAppliedQuery],
+  );
+
+  const onInputChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+    const newQuery = event.target.value;
+
+    setIsActive(false);
+
+    setQuery(newQuery);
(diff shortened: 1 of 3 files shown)


vlevik (approved):
good job 🔥

src/components/Dropdown.tsx
```diff
@@ -0,0 +1,109 @@
...
+import { Person } from '../types/Person';
+
+type Props = {
+  peopleFromServer: Person[];
+  onSelected: (person: Person) => void;
+};
+
+export const Dropdown: React.FC<Props> = ({
+  peopleFromServer,
+  onSelected = () => {},
```
vasylmiskiv, line 17:
```suggestion
  onSelected,
```

src/components/Dropdown.tsx
```diff
@@ -0,0 +1,109 @@
...
+}) => {
+  const [isActive, setIsActive] = useState(false);
+  const [query, setQuery] = useState('');
+  const [appliedQuery, setAppliedQuery] = useState('');
+
+  const applyQuery = useCallback(
+    debounce((value) => {
+      setAppliedQuery(value.trim() !== '' ? value : '');
```
vasylmiskiv, line 25: Tbh I didn't get the logic) If you want to trim your string you could do like so
```suggestion
      setAppliedQuery(value.trim());
```
It will be the same

src/components/Dropdown.tsx
```diff
@@ -0,0 +1,109 @@
...
+      setIsActive(true);
+    }, 1000),
+    [setAppliedQuery],
+  );
+
+  const onInputChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+    const newQuery = event.target.value;
+
+    setIsActive(false);
```
vasylmiskiv, line 34: `setIsActive` calls every time when you typing something. Could you optimize it?

src/components/Dropdown.tsx
```diff
@@ -0,0 +1,109 @@
...
+          onFocus={handleInputFocus}
+          onBlur={handleInputBlur}
+          value={query}
+          onChange={onInputChange}
+        />
+      </div>
+
+      <div className="dropdown-menu" role="menu">
+        <div className="dropdown-content">
+          {suggestions.length > 0
```
vasylmiskiv, line 83:
```suggestion
          {suggestions.length
```

vasylmiskiv (changes requested):
Looks nice but need to add some changes

src/components/Dropdown.tsx
sviatikkondur: can you help me, I tried to use useRef, but it doesn't work?

src/components/Dropdown.tsx
```diff
@@ -0,0 +1,116 @@
...
+    setQuery(person.name);
+    setAppliedQuery(person.name);
+  };
+
+  // setInterval(() => {
+  //   setIsActive(true);
+  //   console.log("Interval working...")
+  // }, 1000);
+
+  // console.log('Rendering...');
```
vasylmiskiv, line 70: Don't forget to remove commented out lines

src/components/Dropdown.tsx
```diff
@@ -0,0 +1,116 @@
...
+      setAppliedQuery(value.trim());
+      setIsActive(true);
+    }, 1000),
+    [setAppliedQuery],
+  );
+
+  const onInputChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+    const newQuery = event.target.value;
+
+    setIsActive(false);
```
vasylmiskiv, line 34: I meant that this function is triggered in any case even if isActive false. You could have just written a check like so and that's it
```suggestion
    if (isActive) {
      setIsActive(false);
    }
   
```

vasylmiskiv (approved):
Well done 🚀

src/components/Dropdown.tsx
sviatikkondur: I get the point now, thanks!

PR #325, opened 2023-10-25
add autocomplete solution

Kuper18:
[DEMO](https://kuper18.github.io/react_autocomplete/)

Hi

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,56 +1,82 @@
-import React from 'react';
+import React, {
+  useCallback,
+  useMemo,
+  useRef,
+  useState,
+} from 'react';
+import debounce from 'lodash.debounce';
 import './App.scss';
 import { peopleFromServer } from './data/people';
+import { Person } from './types/Person';
+import { List } from './List';
 
 export const App: React.FC = () => {
-  const { name, born, died } = peopleFromServer[0];
+  const [isFocusedInput, setIsFocusedInput] = useState(false);
+  const [query, setQuery] = useState('');
+  const [apliedQuery, setApliedQuery] = useState('');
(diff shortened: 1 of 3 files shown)


polosanya (changes requested):
A few moments that would be great to improve:
1. It would be great to make the whole row of the list clickable for better UX
image
2. Why is it not possible to type in upper case? Don't change the text in input, just modify the text on filtering
image

src/App.tsx
```diff
@@ -1,56 +1,82 @@
...
+  const [isFocusedInput, setIsFocusedInput] = useState(false);
+  const [query, setQuery] = useState('');
+  const [apliedQuery, setApliedQuery] = useState('');
+  const [sortedPeople] = useState<Person[]>(peopleFromServer);
+
+  const applyQuery = useCallback(
+    debounce(setApliedQuery, 1000), [],
+  );
+
+  const ref = useRef<Person | null>(null);
```
StasSokolov, line 23: let's set `useRef` before `applyQuery`

src/List.tsx
```diff
@@ -0,0 +1,38 @@
...
+  visiblePeople: Person[];
+  onSelectPerson?: (value: Person) => void;
+};
+
+export const List: React.FC<Props> = ({
+  visiblePeople,
+  onSelectPerson = () => {},
+}) => (
+  <div className="dropdown-menu" role="menu">
+    {visiblePeople.length > 0 ? (
```
StasSokolov, line 14:
```suggestion
    {visiblePeople.length ? (
```
can be simplify

StasSokolov (approved):
Looks good to me, i'm approving it with only few minor comments🔥

PR #330, opened 2023-10-26
add task solution

adioskv:
[DEMO LINK](https://adioskv.github.io/react_autocomplete/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,57 +1,29 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
+
 import { peopleFromServer } from './data/people';
+import { Person } from './types/Person';
+import { Dropdown } from './components/Dropdown';
 
 export const App: React.FC = () => {
-  const { name, born, died } = peopleFromServer[0];
+  const [selectedHuman, setSelectedHuman] = useState<Person | null>(null);
+
+  const onSelect = (person: Person) => (
+    setSelectedHuman(person)
+  );
 
   return (
     <main className="section">
       <h1 className="title">
-        {`${name} (${born} = ${died})`}
+        {selectedHuman
+          ? `${selectedHuman.name} (${selectedHuman.born} = ${selectedHuman.died})`
+          : 'No selected person'}
       </h1>
 
-      <div className="dropdown is-active">
-        <div className="dropdown-trigger">
-          <input
-            type="text"
-            placeholder="Enter a part of the name"
-            className="input"
-          />
-        </div>
-
-        <div className="dropdown-menu" role="menu">
-          <div className="dropdown-content">
-            <div className="dropdown-item">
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -1,57 +1,29 @@
...
 import { peopleFromServer } from './data/people';
+import { Person } from './types/Person';
+import { Dropdown } from './components/Dropdown';
 
 export const App: React.FC = () => {
-  const { name, born, died } = peopleFromServer[0];
+  const [selectedHuman, setSelectedHuman] = useState<Person | null>(null);
+
+  const onSelect = (person: Person) => (
+    setSelectedHuman(person)
```
vasylmiskiv, line 12:
```suggestion
    setSelectedPerson(person)
```

src/components/Dropdown.tsx
```diff
@@ -0,0 +1,97 @@
...
+    if (isActiveSuggestions) {
+      setIsActiveSuggestions(false);
+    }
+
+    setQuery(event.target.value);
+
+    setTimeout(() => {
+      setAppliedQuery(event.target.value);
+      setIsActiveSuggestions(true);
+    }, delay);
```
vasylmiskiv, line 36: It will be triggered every time when user is typing something. Should  use separate debounce function

src/components/Dropdown.tsx
```diff
@@ -0,0 +1,97 @@
...
+      { 'is-active': isActiveSuggestions },
+    )}
+    >
+      <div className="dropdown-trigger">
+        <input
+          type="text"
+          placeholder="Enter a part of the name"
+          className="input"
+          value={query}
+          onChange={(event) => handleQuery(event, 1000)}
```
vasylmiskiv, line 60:
```suggestion
          onChange={handleQueryChange}
```

src/components/Dropdown.tsx
```diff
@@ -0,0 +1,97 @@
...
+                <a
+                  key={human.slug}
+                  className={cn(
+                    'dropdown-item',
+                    {
+                      'has-text-link': human.sex === 'm',
+                      'has-text-danger': human.sex === 'f',
+                    },
+                  )}
+                  onMouseDown={() => handlePersonSelect(human)}
```
vasylmiskiv, line 80: onClick event would be more appropriate here

vasylmiskiv (changes requested):
Looks fine but need to add some changes

src/components/Dropdown.tsx
adioskv: it's not working with `onClick`, so i decided to use `onMouseDown` instead

sintax1s (approved):
Looks fine to me!

StasSokolov (approved):
Looks good to me, i'm approving it🔥

PR #332, opened 2023-10-26
dropdown menu implementation

atrofymovych:


etojeDenys (changes requested):
1. let's hide the dropdown on input unfocus 
2. add the cursor pointer to dropdown items

atrofymovych:
i had to add delay for onBlur because if it happends immediately, it is not possible to choose person

anastasiiavorobiova (changes requested):
The autocomplete doesn't work. It removes a person as a selected person when a user starts typing in the input (it shouldn't) It closes the list of people unexpectedly. Feel free to ask for help in the chat, you can tag me (@Anastasiia Vorobiova) for detailed clarifications

src/App.tsx
```diff
@@ -1,14 +1,63 @@
...
+      }
+    }, hideDelay);
+  };
+
+  function filteredPeople(people: Person[], input: string) {
+    return people.filter(person => person.name
+      .toLowerCase()
+      .includes(input.toLowerCase()));
+  }
+
```
IvanFesenko, line 50: Your riun this function twice - it may cause performance issues
Use variable instead on function to get result .
Note: No need to filter in case query is empty
```suggestion
  const filteredPeople =  people.filter(({name}) => name
      .toLowerCase()
      .includes(input.toLowerCase()))

```

IvanFesenko (changes requested):
Please fix one important issue

anastasiiavorobiova (changes requested):
You have a great progress here! Consider closing the list after selection occurred and onBlur event Screenshot 2023-10-27 at 15 33 36

sTorba24 (approved):
Nicely done!

PR #337, opened 2023-10-28
Dropdown with hooks and delays

ElenaGrasovskaya:
[DEMO LINK](https://ElenaGrasovskaya.github.io/react_autocomplete/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,57 +1,53 @@
-import React from 'react';
+import React, { useCallback, useMemo, useState } from 'react';
 import './App.scss';
 import { peopleFromServer } from './data/people';
+import { Dropdown } from './components/Dropdown';
+import { Person } from './types/Person';
 
 export const App: React.FC = () => {
-  const { name, born, died } = peopleFromServer[0];
+  const [selected, setSelected] = useState<Person | null>(null);
+  const [query, setQuery] = useState('');
+
+  const people = useMemo(() => {
+    if (!query) {
+      return peopleFromServer;
+    }
+
+    return peopleFromServer.filter(
+      (person) => person.name.toLowerCase().includes(query.toLowerCase()),
+    );
+  }, [query]);
+
+  const onQueryChange = useCallback(
+    (event: React.ChangeEvent<HTMLInputElement>) => {
+      setQuery(event.target.value);
+      setSelected(null);
+    },
+    [],
+  );
+
+  const onSelected = useCallback((person: Person | null) => {
+    if (person) {
+      setSelected(person);
+      setQuery('');
+    }
+  }, []);
 
   return (
     <main className="section">
-      <h1 className="title">
-        {`${name} (${born} = ${died})`}
-      </h1>
-
-      <div className="dropdown is-active">
-        <div className="dropdown-trigger">
-          <input
-            type="text"
-            placeholder="Enter a part of the name"
-            className="input"
-          />
-        </div>
-
-        <div className="dropdown-menu" role="menu">
-          <div className="dropdown-content">
-            <div className="dropdown-item">
-              <p className="has-text-link">Pieter Haverbeke</p>
-            </div>
-
-            <div className="dropdown-item">
-              <p className="has-text-link">Pieter Bernard Haverbeke</p>
-            </div>
-
-            <div className="dropdown-item">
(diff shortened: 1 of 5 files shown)


src/App.tsx
```diff
@@ -1,57 +1,66 @@
...
-              <p className="has-text-danger">Elisabeth Hercke</p>
-            </div>
-          </div>
-        </div>
-      </div>
+      {selected ? (
+        <h1 className="title">{`${selected?.name} (${selected?.born} = ${selected?.died})`}</h1>
+      ) : (
+        <h1 className="title">Select a person</h1>
+      )}
```
StasSokolov, line 55: let's write one `h1` tag and only change children

src/components/Dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,108 @@
...
+  delay?: number;
+  selected?: Person | null;
+  onSelected: (person: Person) => void;
+  query: string;
+  onQueryChange: (event: React.ChangeEvent<HTMLInputElement>) => void;
+}
+
+const InnerDropdown: React.FC<Props> = ({
+  people,
+  delay = 150,
```
StasSokolov, line 17: will be good write optional params in the end of the list

src/components/Dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,108 @@
...
+          <div className="dropdown-content">
+            {people.length ? (
+              people.map((person, index) => (
+                <div
+                  role="menuitem"
+                  tabIndex={index}
+                  key={person.slug}
+                  onKeyDown={() => {
+                    onSelected(person);
+                    revealDelay(false);
```
StasSokolov, line 75: don't repeat yourself, create separated function for this logic and reuse

StasSokolov (changes requested):
Almost done, let's improve your code a bit more🔥

1. check please this point
Снимок экрана 2023-10-30 в 11 06 35
Снимок экрана 2023-10-30 в 11 03 18

src/App.tsx
ElenaGrasovskaya: fixed

src/components/Dropdown/Dropdown.tsx
ElenaGrasovskaya: fixed

ElenaGrasovskaya:
I managed to fix my code according to comments given to another student and then fixed mine. :) I have also decreased the time delay for debounce so the changes should be applied intuitively.

superpooperxxx (changes requested):
I think you forgot to update gh pages, cause I see you fixed the delay, but I cannot see it in demo🥲

ElenaGrasovskaya:
To be honest I did renew the deploy 2 days ago and did it once again today... and the dropdown surely reacts much faster... Would appreciate any other suggestions on what else I can do to that :)

superpooperxxx (changes requested):
Hey, looks good, but not exactly according to task
1. Debounce doesn't work. The result should update after some delay when I stop typing, not instantly
2. When there is not matching we should displays `No matching suggestions`, not this
image
3. 
image

ElenaGrasovskaya:
I changed the text in messages and changed the delay on debounce... Seems to work fine, well let me know if it doesn't :)

src/App.tsx
```diff
@@ -1,57 +1,68 @@
...
-
-            <div className="dropdown-item">
-              <p className="has-text-danger">Elisabeth Hercke</p>
-            </div>
-          </div>
-        </div>
-      </div>
+      <Dropdown
+        people={people}
+        delay={150}
```
polosanya, line 60: There is no need to pass the default value

src/components/Dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,107 @@
...
+
+const InnerDropdown: React.FC<Props> = ({
+  people,
+  onSelected,
+  query,
+  onQueryChange,
+  delay = 150,
+  selected,
+}) => {
+  const [reveal, setReveal] = useState(false);
```
polosanya, line 23: image

polosanya (approved):
Well done!
It would be great to add cursor pointer on hover for links
image

PR #340, opened 2023-10-31
Develop

Maksim114488:
https://Maksim114488.github.io/react_autocomplete/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,55 +1,97 @@
-import React from 'react';
+import React, { useState, useRef, useCallback } from 'react';
+import cn from 'classnames';
 import './App.scss';
 import { peopleFromServer } from './data/people';
+import { Person } from './types/Person';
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -1,55 +1,97 @@
...
-
-            <div className="dropdown-item">
-              <p className="has-text-danger">Elisabeth Hercke</p>
-            </div>
-          </div>
+          <label className="dropdown-content">
+            {(query === apliedQuery)
+            && isFocused
+            && filteredPeople.length > 0
+            && (filteredPeople.map(person => (
```
polosanya, line 73: It looks too complicated. Create a variable for this condition or try to simplify it

polosanya (changes requested):
The list doesn't look correct. Please fix the styles
image

polosanya (changes requested):
Comments above

PR #346, opened 2023-11-03
Develop

zvir91:
[DEMO LINK](https://zvir91.github.io/react_autocomplete/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,14 +1,73 @@
-import React from 'react';
+import debounce from 'lodash.debounce';
+import { useMemo, useCallback, useState } from 'react';
 import './App.scss';
 import { peopleFromServer } from './data/people';
+import { Person } from './types/Person';
+import { List } from './components/List/List';
 
 export const App: React.FC = () => {
-  const { name, born, died } = peopleFromServer[0];
+  const [query, setQuery] = useState('');
+  const [appliedQuery, setApplyQuerry] = useState('');
+  const [selectedPerson, setSelectedPerson] = useState('');
+  const [isFocused, setIsFocused] = useState(false);
+
+  // const handleFocus = () => {
+  //   setIsFocused(true);
+  // };
+
+  // const handleBlur = () => {
+  //   setIsFocused(true);
+  // };
+
+  const reset = () => {
+    setSelectedPerson('');
+  };
+
+  const handleValue = () => {
+    let value = query;
+
+    if (selectedPerson) {
+      value = selectedPerson;
+    }
+
+    return value;
+  };
+
+  const onSelected = (newPerson: string) => {
+    if (newPerson !== selectedPerson) {
+      setSelectedPerson(newPerson);
+    }
+  };
+
+  const findPerson = peopleFromServer.find(
+    person => person.name === selectedPerson,
+  );
+
+  // eslint-disable-next-line react-hooks/exhaustive-deps
+  const applyQuerry = useCallback(
+    debounce(setApplyQuerry, 1000),
+    [],
+  );
+
+  const handlerQuery = (event:React.ChangeEvent<HTMLInputElement>) => {
+    setQuery(event.target.value);
+    applyQuerry(event.target.value);
+    reset();
+  };
+
+  const filteredPeople: Person[] = useMemo(() => {
+    return [...peopleFromServer].filter(
+      person => person.name.includes(appliedQuery.trim()),
+    );
+  },
+  [appliedQuery]);
 
   return (
-    <main className="section">
+    <main
+      className="section"
+    >
       <h1 className="title">
-        {`${name} (${born} = ${died})`}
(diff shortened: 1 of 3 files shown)


src/App.tsx
```diff
@@ -1,14 +1,65 @@
...
+
+  const handleValue = () => {
+    let value = query;
+
+    if (selectedPerson) {
+      value = selectedPerson;
+    }
+
+    return value;
+  };
```
Andriy-Shymkiv, line 26: what's the purpose of this function? you returning person or query but in general it's value?? looks redundant and should be refactored

src/App.tsx
```diff
@@ -1,14 +1,65 @@
...
+
+  const onSelected = (newPerson: string) => {
+    if (newPerson !== selectedPerson) {
+      setSelectedPerson(newPerson);
+    }
+  };
```
Andriy-Shymkiv, line 32:
```suggestion
  const handleSelect = (newPerson: string) => {
    if (newPerson !== selectedPerson) {
      setSelectedPerson(newPerson);
    }
  };
```

src/App.tsx
```diff
@@ -1,14 +1,65 @@
...
+
+  const findPerson = peopleFromServer.find(
+    person => person.name === selectedPerson,
+  );
```
Andriy-Shymkiv, line 36: you can save to state a whole person and after that you won't need to find it

src/App.tsx
```diff
@@ -1,14 +1,65 @@
...
+
+  // eslint-disable-next-line react-hooks/exhaustive-deps
```
Andriy-Shymkiv, line 38: this is very important rule, why did you disable it? add proper deps instead!

src/App.tsx
```diff
@@ -1,14 +1,65 @@
...
+    applyQuerry(event.target.value);
+    reset();
+  };
+
+  const filteredPeople: Person[] = useMemo(() => {
+    return [...peopleFromServer].filter(
+      person => person.name.includes(appliedQuery.trim()),
+    );
+  },
+  [appliedQuery]);
```
Andriy-Shymkiv, line 55: please debug this functions because the filtering is not working or working not properly:
image
image

src/App.tsx
```diff
@@ -17,38 +68,35 @@ export const App: React.FC = () => {
...
+              <List
+                person={person}
+                onSelected={onSelected}
+                key={person.name}
+              />
+            ))}
+            {(!selectedPerson && !filteredPeople.find(
+              p => p.name.includes(query),
+            ) && query)
+              && 'No matching suggestions'}
```
Andriy-Shymkiv, line 99: something wrong with conditions because I see this unnecessary white space:
image

src/App.tsx
```diff
@@ -17,38 +68,35 @@ export const App: React.FC = () => {
...
+            {(isFocused && !query)
+            && (!selectedPerson) && peopleFromServer.map(person => (
+              <List
+                person={person}
+                onSelected={onSelected}
+                key={person.name}
+              />
+            ))}
+            {(query && !selectedPerson) && filteredPeople.map((person) => (
+              <List
```
Andriy-Shymkiv, line 90: it is `Person`, not `List`

src/App.tsx
Andriy-Shymkiv: conditions should be refactored 100%

Andriy-Shymkiv (changes requested):

src/App.tsx
zvir91: This function sets value of input(should it be query or selectedPerson)

zvir91: if selectedValue true value of input is SelectedValue else it is query

src/App.tsx
```diff
@@ -17,39 +68,44 @@ export const App: React.FC = () => {
...
+            value={selectedPerson ? selectedPerson.name : query}
+            onChange={handlerQuery}
+            onFocus={() => {
+              setIsFocused(true);
+              setNoSuggestions(false);
+            }}
+            onBlur={() => {
+              setIsFocused(true);
+              setNoSuggestions(false);
+            }}
```
DarkMistyRoom, line 80: looks the same. Create a function and reuse it

src/App.tsx
```diff
@@ -1,14 +1,65 @@
...
+    setSelectedPerson(newPerson);
+  };
+
+  const applyQuerry = useMemo(() => debounce(
+    (value: string) => {
+      setApplyQuerry(value);
+      setNoSuggestions(false);
+    }, 1000,
+  ),
+  [setApplyQuerry, setNoSuggestions]);
```
DarkMistyRoom, line 33: the `delay` should be customizable via props;

src/App.tsx
DarkMistyRoom: consider using a ternary operator instead

DarkMistyRoom (changes requested):
Good job! But hide suggestions on text change

SanyaBratashchuk (approved):

PR #349, opened 2023-11-05
implementation

Daniilart01:
[DEMO LINK](https://Daniilart01.github.io/react_autocomplete/)
In my implementation I use personal implementation of debounce to display loading animation while waiting


src/components/DropDownMenu/DropDownMenu.tsx
```diff
@@ -0,0 +1,43 @@
...
+type Props = {
+  personList: Person[]
+  selectPerson: (person: Person) => void;
+};
+
+export const DropDownMenu: React.FC<Props> = ({ personList, selectPerson }) => {
+  return (
+    <div className="dropdown-menu" role="menu">
+      <div className="dropdown-content">
+        {personList.length !== 0
```
lerastarynets, line 13:
```suggestion
        {!!personList.length
```

src/components/DropDownMenu/DropDownMenu.tsx
```diff
@@ -0,0 +1,43 @@
...
+          ? personList.map(person => (
+            <a
+              href="/#"
```
lerastarynets, line 16:
```suggestion
              href="/"
```
the issue, that you've mentioned is because on click you should also set query (add the name of the person to a browser route)

lerastarynets (changes requested):
Nice start! Let's fix the solution according to the checklist

src/components/DropDownMenu/DropDownMenu.tsx
Daniilart01: Dropdown double negation error occures

OleziO, line 13:
```suggestion
        {personList.length
```

OleziO (changes requested):
Align text to center and fix letter spacing
image

Daniilart01:
> Align text to center and fix letter spacing image

image
It is already aligned and letter-spacing set to default value

lerastarynets (changes requested):
You still have issue with page reload, considering you implementation, I think it's better to just leave href='/' and dob't add person name to browser route, make this work and rerequest review)

Daniilart01:
> You still have issue with page reload, considering you implementation, I think it's better to just leave href='/' and dob't add person name to browser route, make this work and rerequest review)

Done, but in some reason the error occures. I ges will be better to change this links using buttons, but in this way it looks like hardcode

VitaliyBodnar (approved):

PR #355, opened 2023-11-08
solution

Yurko-Komanitskyi:
[DEMO LINK](https://yuron-maker.github.io/react_autocomplete/)

### DIFF
--- a/src/PeopleList.tsx
+++ b/src/PeopleList.tsx
@@ -15,14 +15,19 @@ export const Peoplelist: React.FC<Props> = ({ people, onSelect, delay }) => {
   const [appliedtextField, setAppliedtextField] = useState('');
   const [isHide, setIsHide] = useState(false);
 
+  const handleApplyTextSet = (value: string) => {
+    setAppliedtextField(value);
+    setIsHide(true);
+  };
+
   const applyTextField = useCallback(
-    debounce(setAppliedtextField, delay), [],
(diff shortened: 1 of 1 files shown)


src/PeopleList.tsx
```diff
@@ -0,0 +1,91 @@
...
+
+import { Person } from './types/Person';
+
+interface Props {
+  people: Person[];
+  onSelect: (persone: Person) => void;
+  delay: number;
+}
+
+export const Peoplelist: React.FC<Props> = ({ people, onSelect, delay }) => {
```
anastasiia-tilikina, line 13: In the future try to choose more descriptive component names.
When I see the PeopleList name I (as a developer who's seeing your code at the first time) expect to find only some map of the list in it. But in your case it containes some input and logic for applying query.

anastasiia-tilikina: Ideally it should be some component for instance - Autocomplete that contains input + logic and another component - people list which only accepts currentPeople as a prop and renders it.

anastasiia-tilikina: no need to refactor it now. Just FYI)

anastasiia-tilikina (changes requested):
Well done.

1. hide dropdown on input Blur
2. hide it on input change.

anastasiiavorobiova (approved):
Great work!

PR #356, opened 2023-11-08
version 1.0

HokageHEL:
[DEMO LINK](https://HokageHEL.github.io/react_autocomplete/)

### DIFF
--- a/src/Components/Autocomplete/Autocomplete.tsx
+++ b/src/Components/Autocomplete/Autocomplete.tsx
@@ -9,16 +9,14 @@ import { peopleFromServer } from '../../data/people';
 const toNormalizeQuery = (query: string) => query.trim().toLowerCase();
 
 type Props = {
-  onSelect: (person: Person) => void,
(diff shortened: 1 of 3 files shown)


src/Components/Autocomplete/Autocomplete.tsx
```diff
@@ -0,0 +1,86 @@
...
+          className="input"
+          value={query}
+          onChange={handleQueryChange}
+        />
+      </div>
+
+      {isVisible && (
+        <div className="dropdown-menu" role="menu">
+          <div className="dropdown-content">
+            {filteredPeople.length ? (
```
denys-danyliuk, line 61: when the field looses focus you should hide this list, and in turn when I click on the input would be better to show the list from the beginning
Знімок екрана 2023-11-09 о 10 06 51

denys-danyliuk (changes requested):

denys-danyliuk (approved):
GJ!

PR #362, opened 2023-11-15
Add tast solution

ANTI-bogdan:
[DEMO LINK](https://ANTI-bogdan.github.io/react_autocomplete/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,57 +1,26 @@
-import React from 'react';
+import React, { useCallback, useState } from 'react';
+
 import './App.scss';
 import { peopleFromServer } from './data/people';
+import { Person } from './types/Person';
+import { Autocomplete } from './components/Autocompleted';
 
 export const App: React.FC = () => {
-  const { name, born, died } = peopleFromServer[0];
+  const [selectedPerson, setSelectedPerson] = useState<Person | null>(null);
 
   return (
     <main className="section">
       <h1 className="title">
-        {`${name} (${born} = ${died})`}
+        {selectedPerson
+          ? `${selectedPerson.name} ${selectedPerson.born} - ${selectedPerson.died}`
+          : 'No selected person'}
       </h1>
 
-      <div className="dropdown is-active">
-        <div className="dropdown-trigger">
-          <input
-            type="text"
-            placeholder="Enter a part of the name"
-            className="input"
-          />
-        </div>
-
-        <div className="dropdown-menu" role="menu">
-          <div className="dropdown-content">
-            <div className="dropdown-item">
-              <p className="has-text-link">Pieter Haverbeke</p>
-            </div>
-
-            <div className="dropdown-item">
-              <p className="has-text-link">Pieter Bernard Haverbeke</p>
-            </div>
-
-            <div className="dropdown-item">
-              <p className="has-text-link">Pieter Antone Haverbeke</p>
-            </div>
-
-            <div className="dropdown-item">
-              <p className="has-text-danger">Elisabeth Haverbeke</p>
-            </div>
-
-            <div className="dropdown-item">
-              <p className="has-text-link">Pieter de Decker</p>
-            </div>
-
-            <div className="dropdown-item">
-              <p className="has-text-danger">Petronella de Decker</p>
-            </div>
-
-            <div className="dropdown-item">
-              <p className="has-text-danger">Elisabeth Hercke</p>
-            </div>
-          </div>
-        </div>
-      </div>
+      <Autocomplete
+        people={peopleFromServer}
+        timer={1000}
+        onSelect={setSelectedPerson}
+      />
     </main>
   );
 };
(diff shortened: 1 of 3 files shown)


src/App.tsx
```diff
@@ -1,57 +1,26 @@
...
 export const App: React.FC = () => {
-  const { name, born, died } = peopleFromServer[0];
+  const [selectedPerson, setSelectedPerson] = useState<Person | null>(null);
 
   return (
     <main className="section">
       <h1 className="title">
-        {`${name} (${born} = ${died})`}
+        {selectedPerson
+          ? `${selectedPerson.name} ${selectedPerson.born} - ${selectedPerson.died}`
```
volodymyr-soltys97, line 15: Move this logic from jsx to the helper variable and use it here

src/components/Autocompleted.tsx
```diff
@@ -0,0 +1,89 @@
...
+          {filteredPeople.length
+            ? filteredPeople.map(person => (
+              <a
+                href="/#"
+                className="dropdown-item"
+                key={person.name}
+                onMouseDown={() => handlePersonSelect(person)}
+              >
+                <p className={classNames({
+                  'has-text-link': person.sex === 'm',
```
volodymyr-soltys97, line 72: Consider will be better if you create a variable for this magic words and use it everywhere

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code
1. When I select a person and then clear the field the person's name stays at the top, it would be good if it clears as well 
image

2. Also try to fix the problem with tests, try to write this command in the terminal
image

src/components/Autocompleted.tsx
ANTI-bogdan: why?

src/App.tsx
ANTI-bogdan: why?

volodymyr-soltys97: It's a bad practice write logic in the jsx, it's make your code more readable

src/components/Autocompleted.tsx
volodymyr-soltys97: This will make the code more readable, for example if you need to write it in many places, and then change the condition to another, then in order not to change it everywhere, it is better to use a variable or an enum for strings

src/components/Autocompleted.tsx
```diff
@@ -0,0 +1,97 @@
...
+                  'has-text-danger': person.sex === 'f',
+                })}
+                >
+                  {person.name}
```
volodymyr-soltys97, line 84: Also you can use a destruction for `person`

volodymyr-soltys97 (changes requested):
Let's improve your code

src/components/Autocompleted.tsx
```diff
@@ -0,0 +1,103 @@
...
+          onFocus={() => setIsFocused(true)}
+          onBlur={() => setIsFocused(false)}
+          onChange={handleQueryChange}
+        />
+      </div>
+
+      <div className="dropdown-menu" role="menu">
+        <div className="dropdown-content">
+          {filteredPeople.length
+            ? filteredPeople.map(person => {
```
VitaliyBodnar, line 72:
```suggestion
          {filteredPeople?.map(person => {
```

src/components/Autocompleted.tsx
```diff
@@ -0,0 +1,103 @@
...
+    })}
+    >
+      <div className="dropdown-trigger">
+        <input
+          type="text"
+          placeholder="Enter a part of the name"
+          className="input"
+          value={query}
+          onFocus={() => setIsFocused(true)}
+          onBlur={() => setIsFocused(false)}
```
VitaliyBodnar, line 64: It's a bad practice to create anonymous function is jsx, it's better to create separate function for that

VitaliyBodnar: Check other places

VitaliyBodnar (changes requested):

src/components/Autocompleted.tsx
ANTI-bogdan: it`s not work

etojeDenys (changes requested):
1. you forgot to implement this option
image

etojeDenys (changes requested):
another issue appears, the selected person should be a value in the input
image

src/App.tsx
```diff
@@ -1,57 +1,28 @@
...
+  const [selectedPerson, setSelectedPerson] = useState<Person | null>(null);
+  const { name, born, died } = selectedPerson || {};
 
   return (
     <main className="section">
       <h1 className="title">
-        {`${name} (${born} = ${died})`}
+        {selectedPerson ? (
+          `${name} (${born} - ${died})`
+        ) : 'No selected person'}
```
volodymyr-soltys97, line 17: You can create a variable for this checking and use it here

volodymyr-soltys97 (approved):
Great work 🔥

PR #371, opened 2023-11-20
Develop

Serhii-Shtokalo:
- [DEMO LINK](https://Serhii-Shtokalo.github.io/react_autocomplete/)


loralevitska (changes requested):
Re-deploy your solution and check demo link before re-request review
image

PR #373, opened 2023-11-21
Develop

vladkorobka:
DEMO
https://vladkorobka.github.io/react_autocomplete/


volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your page.
Consider will be better, if you will clear the text above after we delete the person in the input
image

anastasiiavorobiova (changes requested):
Great work! Consider showing a message if there are no matching suggestions

anastasiiavorobiova (approved):
Awesome work!

PR #376, opened 2023-11-22
add task solution

KolomietsN:
[DEMO LINK](https://KolomietsN.github.io/react_autocomplete/)


Esceype (changes requested):
GJ!
Need to fix:
- if type something and then delete it before the loader disappear then the loader will spinning infinitely;
- while user type list should be empty;

etojeDenys (changes requested):
looks good, just implement this option from the task description
image

PR #381, opened 2023-11-25
solution

ArtemSydun:
review pls, i don`t know how to optimize component with useMemo and stop person select redirect to another page (it`s not redirecting whenever on live server)
[DEMO](https://artemsydun.github.io/react_autocomplete/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,56 +1,56 @@
-import React from 'react';
+import React, { useState, useMemo } from 'react';
+import cn from 'classnames';
 import './App.scss';
 import { peopleFromServer } from './data/people';
+import { Person } from './types/Person';
+import { PeopleList } from './components/PeopleList';
 
 export const App: React.FC = () => {
-  const { name, born, died } = peopleFromServer[0];
+  const [isFocused, setIsFocused] = useState(false);
+  const [inputQuery, setInputQuery] = useState('');
+  const [selectedPerson, setSelectedPerson] = useState<Person | null>(null);
+
+  const filterQuery = useMemo(() => {
+    return peopleFromServer.filter(person => person.name.toLowerCase()
+      .includes(inputQuery.toLowerCase()));
+  }, [inputQuery]);
+
+  const onSelect = (person: Person) => {
+    setInputQuery(person.name);
+    setSelectedPerson(person);
+    setIsFocused(false);
+  };
 
   return (
     <main className="section">
       <h1 className="title">
-        {`${name} (${born} = ${died})`}
+        {selectedPerson ? (
+          `${selectedPerson.name} (${selectedPerson.born} - ${selectedPerson.died})`
+        ) : (
+          'No matches found'
+        )}
+
       </h1>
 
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -1,56 +1,56 @@
...
+
+  const filterQuery = useMemo(() => {
+    return peopleFromServer.filter(person => person.name.toLowerCase()
+      .includes(inputQuery.toLowerCase()));
+  }, [inputQuery]);
+
+  const onSelect = (person: Person) => {
+    setInputQuery(person.name);
+    setSelectedPerson(person);
+    setIsFocused(false);
```
artemforsoff, line 21: the input has a special event when we can understand that the focus is lost

src/components/PeopleList.tsx
```diff
@@ -0,0 +1,39 @@
...
+
+export const PeopleList: React.FC<Props> = React.memo((
+  { personList, onSelect },
+) => {
+  return (
+    <div className="dropdown-menu" role="menu">
+      <div className="dropdown-content">
+        {personList.length ? (personList.map(person => (
+          // eslint-disable-next-line jsx-a11y/click-events-have-key-events, jsx-a11y/no-static-element-interactions, jsx-a11y/anchor-is-valid
+          <a
```
artemforsoff, line 18: if a click should cause a redirect, then we use a link tag, otherwise it’s a button tag

artemforsoff (changes requested):
overall the solution is very similar to what it should be 🤟🏻
let's look again at what is required in the task
If something is not clear, I’ll wait in chat and be happy to answer

src/App.tsx
```diff
@@ -1,56 +1,56 @@
...
+  };
 
   return (
     <main className="section">
       <h1 className="title">
-        {`${name} (${born} = ${died})`}
+        {selectedPerson ? (
+          `${selectedPerson.name} (${selectedPerson.born} - ${selectedPerson.died})`
+        ) : (
+          'No matches found'
```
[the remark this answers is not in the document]
ArtemSydun: you mean change 'no mathces'; on 'no selected person'?

BudnikOleksii (changes requested):
Good work, but a lot of comments are not fixed.

src/components/PeopleList.tsx
ArtemSydun: as i see other pull requests approved with <a> tag, so what should i do? replace it with button tag?

yevtok (changes requested):
When you lost focus on input List must disappear

src/components/Autocomplete.tsx
```diff
@@ -0,0 +1,100 @@
...
+
+interface Props {
+  personList: Person[];
+  onSelect?: (person: Person) => void;
+  delay: number;
+}
+
+export const Autocomplete: React.FC<Props> = ({
+  personList,
+  onSelect = () => { },
```
roman-mirzoian, line 14:
```suggestion
  onSelect = () => {},
```

PR #391, opened 2023-12-02
made solution to autocomplete

MrHunterBoi:
[DEMO LINK](https://MrHunterBoi.github.io/react_autocomplete/)


ArtemSydun (approved):
gj

src/helpers.ts
```diff
@@ -0,0 +1,12 @@
+// eslint-disable-next-line @typescript-eslint/ban-types
```
roman-mirzoian, line 1: Try not to disable the linter, but to describe everything that is needed so that it does not quarrel

PR #396, opened 2023-12-05
add task solution

ruslanagnatenko:
[DEMO LINK](https://ruslanagnatenko.github.io/react_autocomplete/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,56 +1,104 @@
-import React from 'react';
+import React, { useCallback, useMemo, useState } from 'react';
+import cn from 'classnames';
+import debounce from 'lodash.debounce';
+
 import './App.scss';
 import { peopleFromServer } from './data/people';
+import { Person } from './types/Person';
 
 export const App: React.FC = () => {
-  const { name, born, died } = peopleFromServer[0];
+  const [isFocused, setIsFocused] = useState(false);
+  const [query, setQuery] = useState('');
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -1,56 +1,104 @@
...
+
 import './App.scss';
 import { peopleFromServer } from './data/people';
+import { Person } from './types/Person';
 
 export const App: React.FC = () => {
-  const { name, born, died } = peopleFromServer[0];
+  const [isFocused, setIsFocused] = useState(false);
+  const [query, setQuery] = useState('');
+  const [appliedQuery, setAppliedQuery] = useState('');
```
polosanya, line 12: Instead of having two states, you can just filter the people on debounce

src/App.tsx
```diff
@@ -1,56 +1,104 @@
...
-        {`${name} (${born} = ${died})`}
-      </h1>
+      {selectedPerson
+        ? (
+          <h1 className="title">
+            {`${selectedPerson.name} (${selectedPerson.born} = ${selectedPerson.died})`}
+          </h1>
+        )
+        : (
+          <h1 className="title">No selected person</h1>
```
polosanya, line 49: Ternary operator can be placed inside h1 tag as it's the same for both cases

polosanya (changes requested):
Almost done! Try to fix eslint problems

PR #400, opened 2023-12-06
optional task solution

VeDono:
[DEMO LINK](https://VeDono.github.io/react_autocomplete/)

### DIFF
--- a/src/components/Autocomplete.tsx
+++ b/src/components/Autocomplete.tsx
@@ -1,0 +1,108 @@
+import React, { useMemo, useState } from 'react';
+import cn from 'classnames';
+import debounce from 'lodash.debounce';
+
+import { Person } from '../types/Person';
+
+type Props = {
+  peopleFromServer: Person[];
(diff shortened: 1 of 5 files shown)


src/components/Autocomplete.tsx
```diff
@@ -0,0 +1,108 @@
...
+  }, [appliedQuery, peopleFromServer]);
+
+  const applySearchQuery = useMemo(() => {
+    return debounce(setAppliedQuery, 1000);
+  }, []);
+
+  const handleChangeInput = (event: React.ChangeEvent<HTMLInputElement>) => {
+    setSearchQuery(event.target.value);
+    applySearchQuery(event.target.value);
+  };
```
BudnikOleksii, line 57:
```suggestion
  const handleChangeInput = (event: React.ChangeEvent<HTMLInputElement>) => {
    const { value } = event.target;
  
    setSearchQuery(value);
    applySearchQuery(value);
  };
```
Use destructing to avoid repeating

PR #405, opened 2023-12-09
Created autocomplete

GeorgeSavostikov:
[DEMO LINK](https://GeorgeSavostikov.github.io/react_autocomplete/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,57 +1,27 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
 import { peopleFromServer } from './data/people';
+import { Person } from './types/Person';
+import { Autocomplete } from './components/Autocomplete';
 
 export const App: React.FC = () => {
-  const { name, born, died } = peopleFromServer[0];
+  const [myPerson, setMyPerson] = useState<Person | null>(null);
 
   return (
     <main className="section">
       <h1 className="title">
-        {`${name} (${born} = ${died})`}
+        {myPerson ? (
+          `${myPerson.name} (${myPerson.born} - ${myPerson.died})`
+        ) : (
+          'No selected person'
+        )}
       </h1>
 
-      <div className="dropdown is-active">
-        <div className="dropdown-trigger">
-          <input
-            type="text"
-            placeholder="Enter a part of the name"
-            className="input"
-          />
-        </div>
-
-        <div className="dropdown-menu" role="menu">
-          <div className="dropdown-content">
-            <div className="dropdown-item">
-              <p className="has-text-link">Pieter Haverbeke</p>
-            </div>
-
-            <div className="dropdown-item">
-              <p className="has-text-link">Pieter Bernard Haverbeke</p>
-            </div>
-
-            <div className="dropdown-item">
-              <p className="has-text-link">Pieter Antone Haverbeke</p>
-            </div>
-
-            <div className="dropdown-item">
-              <p className="has-text-danger">Elisabeth Haverbeke</p>
-            </div>
-
-            <div className="dropdown-item">
-              <p className="has-text-link">Pieter de Decker</p>
-            </div>
-
-            <div className="dropdown-item">
-              <p className="has-text-danger">Petronella de Decker</p>
-            </div>
-
-            <div className="dropdown-item">
-              <p className="has-text-danger">Elisabeth Hercke</p>
-            </div>
-          </div>
-        </div>
-      </div>
+      <Autocomplete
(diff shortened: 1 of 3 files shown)


src/components/Autocomplete.tsx
```diff
@@ -0,0 +1,102 @@
+import React, { useState, useMemo } from 'react';
+import cn from 'classnames';
+import debounce from 'lodash.debounce';
```
SanyaBratashchuk, line 3: add it to the project dependencies specified in package.json

src/components/Autocomplete.tsx
```diff
@@ -0,0 +1,102 @@
...
+  const [select, setSelect] = useState(false);
+  const [query, setQuery] = useState('');
+  const [chosePerson, setChosePerson] = useState('');
+
+  const applyQuery = useMemo(
+    () => debounce(setQuery, delay), [delay],
+  );
+
+  const filteredPeoples = useMemo(() => {
+    let newArr = [...peopleFromServer];
```
SanyaBratashchuk, line 26: you don't need to make a copy since filter returns new array

SanyaBratashchuk (changes requested):

src/App.tsx
```diff
@@ -1,57 +1,27 @@
...
-
-            <div className="dropdown-item">
-              <p className="has-text-danger">Elisabeth Hercke</p>
-            </div>
-          </div>
-        </div>
-      </div>
+      <Autocomplete
+        peopleFromServer={peopleFromServer}
+        setMyPerson={(person) => setMyPerson(person)}
```
etojeDenys, line 22:
```suggestion
        setMyPerson={setMyPerson}
```
we can simplify it

etojeDenys (changes requested):
looks like you forgot to implement this option
image

src/components/Autocomplete.tsx
```diff
@@ -0,0 +1,101 @@
...
+          type="text"
+          placeholder="Enter a part of the name"
+          className="input"
+          value={chosePerson}
+          onChange={handleChangeQuery}
+          onFocus={() => setSelect(true)}
+          onBlur={() => {
+            setSelect(false);
+            setQuery('');
+          }}
```
VitaliyBodnar, line 68: it is bad practise to make anonymous arrow functions in jsx.
make handler in component and pass reference here

src/components/Autocomplete.tsx
```diff
@@ -0,0 +1,101 @@
...
+    if (query !== '') {
+      return peopleFromServer.filter(
+        (person: Person) => person.name
+          .toLowerCase()
+          .includes(query.toLowerCase()),
+      );
+    }
+
+    return peopleFromServer;
+  }, [peopleFromServer, query]);
```
VitaliyBodnar, line 35: let's make a code a little bit cleaner 
```suggestion
  const filteredPeoples = useMemo(() => {
    if (!query) {
        return peopleFromServer;
    }
    
    return peopleFromServer.filter(
       (person: Person) => person.name
          .toLowerCase()
          .includes(query.toLowerCase())
    );
  }, [peopleFromServer, query]);
```

VitaliyBodnar (changes requested):

src/App.tsx
```diff
@@ -1,57 +1,27 @@
...
 export const App: React.FC = () => {
-  const { name, born, died } = peopleFromServer[0];
+  const [myPerson, setMyPerson] = useState<Person | null>(null);
 
   return (
     <main className="section">
       <h1 className="title">
-        {`${name} (${born} = ${died})`}
+        {myPerson ? (
+          `${myPerson.name} (${myPerson.born} - ${myPerson.died})`
```
volodymyr-soltys97, line 14: Consider will be better if you move this to the helper variable and use it here

src/components/Autocomplete.tsx
```diff
@@ -0,0 +1,106 @@
...
+            filteredPeoples.map((person: Person) => (
+              <button
+                type="button"
+                className="dropdown-item button is-white"
+                key={person.slug}
+                onMouseDown={() => chosenPerson(person)}
+              >
+                <p
+                  className={cn({
+                    'has-text-link': person.sex === 'm',
```
volodymyr-soltys97, line 89: You can create a enum for this magic words 'm' and 'f' and use it here

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code
It will be better if you also clean up the text above after we have removed the person in the input
image

volodymyr-soltys97 (approved):
Well done 🔥

PR #409, opened 2023-12-11
solution 1

mickhakk:
[DEMO LINK](https://mickhakk.github.io/react_autocomplete/)

### DIFF
--- a/src/components/People/People.tsx
+++ b/src/components/People/People.tsx
@@ -1,0 +1,33 @@
+import { Person } from '../../types/Person';
+
+interface Props {
+  onSelected: (person: Person) => void;
+  peopleToView: Person[];
+}
+
+export function People({ peopleToView, onSelected }: Props) {
+  return (
+    <div className="dropdown-menu" role="menu">
+      <div className="dropdown-content">
+        {peopleToView.length === 0 ? (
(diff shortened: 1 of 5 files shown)


MrHunterBoi (approved):
Good job

src/components/People/People.tsx
```diff
@@ -0,0 +1,33 @@
...
+          </div>
+        ) : (
+          peopleToView.map((person: Person) => (
+            <div
+              aria-hidden
+              className="dropdown-item"
+              key={person.slug}
+              onClick={() => onSelected(person)}
+            >
+              <p className={`${person.sex === 'm' ? 'has-text-link' : 'has-text-danger'}`}>
```
roman-mirzoian, line 24: Avoid using "magic values"

roman-mirzoian (changes requested):
At the moment, technically everything is done correctly and it would be cool to think about the UI part (user experience) as well.
Everything that needs to be clicked should be done with a hover effect and highlighted so that the user understands what he needs to do without prompts. It would also be cool to highlight the current selection so the user knows where it is.
Here's an example of what it might look like:
https://gyazo.com/e7bc1fbe5606509b603b04ea8fe6a1c8

PR #417, opened 2023-12-17
Develop

vinogradova8:
[DEMO LINK](https://vinogradova8.github.io/react_autocomplete/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,56 +1,108 @@
-import React from 'react';
+import cn from 'classnames';
+import React, { useCallback, useMemo, useState } from 'react';
+import debounce from 'lodash.debounce';
 import './App.scss';
 import { peopleFromServer } from './data/people';
 
-export const App: React.FC = () => {
-  const { name, born, died } = peopleFromServer[0];
+type Person = {
+  name: string,
+  sex: string,
+  born: number,
+  died: number,
+  fatherName: string | null,
+  motherName: string | null,
+  slug: string,
+};
+
+type Props = {
+  delay: number,
+};
+
+export const App: React.FC<Props> = ({ delay = 1000 }) => {
+  // const { name, born, died } = peopleFromServer[0];
+
+  const [query, setQuery] = useState('');
+
+  const [appliedQuery, setAppliedQuery] = useState('');
+
+  const [selectedPerson, setSelectedPerson] = useState<Person | null>(null);
+
+  const [isFocused, setIsFocused] = useState(false);
+
+  const applyQuery = useCallback(
+    debounce(setAppliedQuery, delay),
+    [],
+  );
+
+  const handleQueryChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+    setQuery(event.target.value);
+    applyQuery(event.target.value);
+  };
+
+  const filteredPeople = useMemo(() => {
+    return peopleFromServer
+      .filter(personFromServer => personFromServer
+        .name.toLowerCase().includes(appliedQuery.toLowerCase()));
+  }, [appliedQuery]);
 
   return (
     <main className="section">
-      <h1 className="title">
-        {`${name} (${born} = ${died})`}
-      </h1>
-
-      <div className="dropdown is-active">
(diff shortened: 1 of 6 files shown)


src/App.tsx
```diff
@@ -1,56 +1,106 @@
...
+                    className="dropdown-item"
+                    key={person.name}
+                  >
+                    <button
+                      type="button"
+                      className="has-text-link"
+                      onClick={() => {
+                        setSelectedPerson(person);
+                        setQuery(person.name);
+                      }}
```
volodymyr-soltys97, line 90: Move this logic from jsx to the helper function and use it here

src/App.tsx
```diff
@@ -1,56 +1,106 @@
...
-              <p className="has-text-danger">Elisabeth Hercke</p>
+        { query && !appliedQuery ? (
+          <div />
+        ) : (
+          <div className="dropdown-menu" role="menu">
+            <div
+              className="dropdown-content"
+            >
+              {filteredPeople.length !== 0 ? (
+                filteredPeople.map((person) => (
```
volodymyr-soltys97, line 79: Use a destruction for `person`

src/App.tsx
```diff
@@ -1,56 +1,106 @@
...
-            <div className="dropdown-item">
-              <p className="has-text-danger">Elisabeth Hercke</p>
+        { query && !appliedQuery ? (
+          <div />
+        ) : (
+          <div className="dropdown-menu" role="menu">
+            <div
+              className="dropdown-content"
+            >
+              {filteredPeople.length !== 0 ? (
```
volodymyr-soltys97, line 78:
```suggestion
              {!!filteredPeople.length ? (
```

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code
The text above also needs to be cleaned up after we remove the person manually in the input
image

src/App.tsx
```diff
@@ -1,56 +1,108 @@
...
+  const [appliedQuery, setAppliedQuery] = useState('');
+
+  const [selectedPerson, setSelectedPerson] = useState<Person | null>(null);
+
+  const [isFocused, setIsFocused] = useState(false);
+
+  const applyQuery = useCallback(
+    debounce(setAppliedQuery, delay),
+    [],
+  );
```
volodymyr-soltys97, line 33: Need to fix it
image

src/App.tsx
```diff
@@ -1,56 +1,108 @@
...
   return (
     <main className="section">
-      <h1 className="title">
-        {`${name} (${born} = ${died})`}
-      </h1>
-
-      <div className="dropdown is-active">
+      {selectedPerson && query === selectedPerson.name ? (
+        <h1 className="title">
+          {`${selectedPerson.name} (${selectedPerson.born} - ${selectedPerson.died})`}
```
volodymyr-soltys97, line 55: Will be better if you move this logic to the helper variable and use it here

src/App.tsx
```diff
@@ -1,56 +1,108 @@
...
+        </h1>
+      ) : (
+        <h1 className="title">
+          No selected person
+        </h1>
+      )}
```
volodymyr-soltys97, line 61: Also, in order not to duplicate the same tag with the same class twice, it would be better to move this check inside the tag

src/index.tsx
```diff
@@ -2,4 +2,6 @@ import ReactDOM from 'react-dom';
 import 'bulma/css/bulma.css';
 import { App } from './App';
 
-ReactDOM.render(<App />, document.getElementById('root'));
+const delay = 1000;
+
+ReactDOM.render(<App delay={delay} />, document.getElementById('root'));
```
volodymyr-soltys97, line 7:
```suggestion
const delay = 1000;

ReactDOM.render(<App delay={delay} />, document.getElementById('root'));
```
```suggestion
ReactDOM.render(<App />, document.getElementById('root'));
```

volodymyr-soltys97 (changes requested):
Let's improve your code

xiimiik (approved):
GJ

PR #428, opened 2023-12-26
optimization is not working

andrii-rykal:
[DEMO LINK](https://andrii-rykal.github.io/react_autocomplete/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,57 +1,52 @@
-import React from 'react';
+import React, { useCallback, useMemo, useState } from 'react';
+import debounce from 'lodash.debounce';
 import './App.scss';
 import { peopleFromServer } from './data/people';
+import { Autocomplete } from './components/Autocomplete';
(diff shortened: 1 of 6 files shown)


etojeDenys (changes requested):
- the delay should be customizable via props;
- hide suggestions on text change;
- save selected suggestion text to the input on click and close the list;
you forgot to implement these options

src/App.tsx
```diff
@@ -1,57 +1,33 @@
...
-
-            <div className="dropdown-item">
-              <p className="has-text-danger">Elisabeth Hercke</p>
-            </div>
-          </div>
-        </div>
-      </div>
+      <Autocomplete
+        people={filteredPeople}
+        dalay={1000}
```
Viktor-Kost, line 27: Decrease this value. 0.5 sec is enough

Viktor-Kost (approved):

PR #435, opened 2024-01-08
Develop

boytrend0108:
[DEMO LINK](https://boytrend0108.github.io/react_autocomplete/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,9 +1,12 @@
-import React from 'react';
+import { useState } from 'react';
 import './App.scss';
 import { peopleFromServer } from './data/people';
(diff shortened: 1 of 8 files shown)


src/App.tsx
```diff
@@ -1,57 +1,25 @@
...
-
-            <div className="dropdown-item">
-              <p className="has-text-danger">Elisabeth Hercke</p>
-            </div>
-          </div>
-        </div>
-      </div>
+      {person ? (
+        <h1 className="title">
+          {`${person.name} (${person.born} = ${person.died})`}
```
volodymyr-soltys97, line 13: You can create this logic to the variable and use it here

PR #439, opened 2024-01-09
solution

MykhailoLoniak:
[DEMO LINK](https://MykhailoLoniak.github.io/react_autocomplete/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,9 +1,46 @@
-import React from 'react';
-import './App.scss';
+import React, { useCallback, useMemo, useState } from 'react';
+import debounce from 'lodash.debounce';
+
 import { peopleFromServer } from './data/people';
 
+import './App.scss';
+// import { Person } from './types/Person';
+
 export const App: React.FC = () => {
+  const [change, setChange] = useState('');
+
   const { name, born, died } = peopleFromServer[0];
+  const [appliedValue, setAppliedValue] = useState('');
+  const [isDropdownActive, setIsDropdownActive] = useState(false);
+
+  const handleDebouncedChange = useCallback(debounce((inputValue: string) => {
+    setAppliedValue(inputValue);
+  }, 1000), []);
+
+  const handleInputFocus = () => {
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -1,54 +1,95 @@
-import React from 'react';
-import './App.scss';
+import React, { useCallback, useMemo, useState } from 'react';
+import debounce from 'lodash.debounce';
+
 import { peopleFromServer } from './data/people';
 
+import './App.scss';
+// import { Person } from './types/Person';
```
loralevitska, line 7: Remove comment. Try to write obvious code with meaningful variable and function naming

src/App.tsx
```diff
@@ -1,54 +1,95 @@
...
+  }, []);
 
   return (
     <main className="section">
       <h1 className="title">
         {`${name} (${born} = ${died})`}
       </h1>
 
-      <div className="dropdown is-active">
+      <div className={`dropdown ${isDropdownActive ? 'is-active' : ''}`}>
```
loralevitska, line 51: Use [classnames](https://www.npmjs.com/package/classnames) library to add classes conditionally

loralevitska (changes requested):
Almost done)

Add string.trim() for search 
image
Fix error (check in which dependency this library, is it work for you locally)
image
If you have any questions, please ask them in fe_chat and tag mentors. It will be faster, because your answers in GitHub will be visible to mentor only after you re-request review

src/App.tsx
```diff
@@ -1,57 +1,113 @@
-import React from 'react';
-import './App.scss';
+import React, { useCallback, useMemo, useState } from 'react';
+import debounce from 'lodash.debounce';
```
volodymyr-soltys97, line 2: Still not fixed from the previous review
image

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code
We need to clear the text above after we clear the person in the input
image

IvanFesenko (approved):
Well done

PR #440, opened 2024-01-09
react with ts

nazardudnyk:
-- [DEMO LINK](https://nazardudnyk.github.io/react_autocomplete/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,57 +1,38 @@
-import React from 'react';
+import React, { useState, useEffect } from 'react';
 import './App.scss';
 import { peopleFromServer } from './data/people';
+import { Autocomplete } from './Autocomplete';
+import { Person } from './types/Person';
 
 export const App: React.FC = () => {
-  const { name, born, died } = peopleFromServer[0];
+  const [selectedPerson, setSelectedPerson] = useState<Person | null>(null);
+
+  const handleSelectedPerson = (person: Person | null) => {
+    setSelectedPerson(person);
+  };
+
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -1,57 +1,38 @@
...
+
+  const headerText = selectedPerson
+    ? `${selectedPerson.name} (${selectedPerson.born} = ${selectedPerson.died})`
+    : 'No selected person';
+
+  useEffect(() => {
+    if (!selectedPerson) {
+      handleSelectedPerson(null);
+    }
+  }, [selectedPerson]);
```
TarasHoliuk, line 23: this useEffect doesn't make sense - if `selectedPerson === null` (it could be null or Person object) you are trying to set it `null` again

remove it

src/Autocomplete.tsx
```diff
@@ -0,0 +1,92 @@
...
+  const handleInputChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+    setIsDropdownOpen(false);
+    setQuery(event.target.value);
+    applyQuery(event.target.value);
+    onSelected(null);
+  };
+
+  if (clearSelected) {
+    onSelected(null);
+  }
```
TarasHoliuk, line 53: why do you need it?

TarasHoliuk (changes requested):
Very good job :+1: But there are 2 points I cannot get :) I guess you can remove them

SanyaBratashchuk (approved):

PR #442, opened 2024-01-10
solution

SvitlanaRamanauskas:
[DEMO LINK](https://SvitlanaRamanauskas.github.io/react_autocomplete/)


IvanFesenko (changes requested):
Looks good, but what is that? Looks like list should be hidden
2024-01-10_15h42_35

etojeDenys (changes requested):
1. let's hide the dropdown when input is unfocused
2. you forgot to implement this option
image

IvanFesenko (approved):
Well done

PR #444, opened 2024-01-15
Solution

SashaViun:
[DEMO LINK](https://SashaViun.github.io/react_autocomplete/)


etojeDenys (changes requested):
1. save the selected name in the input 
image

2. 
image

DarkMistyRoom (changes requested):
There is no possibility of typing or deleting text in the input

PR #448, opened 2024-01-16
solution

afesh4ak:
[DEMO LINK](https://afesh4ak.github.io/react_autocomplete/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,54 +1,89 @@
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import './App.scss';
 import { peopleFromServer } from './data/people';
+import { Person } from './types/Person';
 
 export const App: React.FC = () => {
-  const { name, born, died } = peopleFromServer[0];
+  const [inputText, setInputText] = useState('');
+  const [suggestions, setSuggestions] = useState<Person[]>([]);
+  const [selectedPerson, setSelectedPerson] = useState<Person | null>(null);
+
+  const handleInputChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+    const newText = event.target.value;
+
+    setInputText(newText);
+
+    setSelectedPerson(null);
+  };
+
+  useEffect(() => {
+    const debounceTimeout = setTimeout(() => {
+      const filteredSuggestions: Person[] = peopleFromServer.filter(
+        (person) => person.name.toLowerCase().includes(inputText.toLowerCase()),
+      );
+
+      setSuggestions(filteredSuggestions);
+    }, 1000);
+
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -1,54 +1,89 @@
...
+    setInputText(person.name);
+  };
 
   return (
     <main className="section">
       <h1 className="title">
-        {`${name} (${born} = ${died})`}
+        {selectedPerson
+          ? `${selectedPerson.name} (${selectedPerson.born} - ${selectedPerson.died})`
+          : 'No selected person'}
```
volodymyr-soltys97, line 41: Move this logic to the variable and use it here

src/App.tsx
```diff
@@ -1,54 +1,89 @@
...
-
-            <div className="dropdown-item">
-              <p className="has-text-danger">Petronella de Decker</p>
-            </div>
+            {inputText === '' && (
+              <div className="dropdown-item">
+                <p className="has-text-link">Show all people</p>
+              </div>
+            )}
+            {suggestions.length === 0 && inputText !== '' && (
```
volodymyr-soltys97, line 62:
```suggestion
            {!suggestions.length && inputText && (
```

src/App.tsx
```diff
@@ -1,54 +1,89 @@
...
+                key={person.name}
+                className="dropdown-item"
+                role="option"
+                aria-selected={person === selectedPerson ? 'true' : 'false'}
+                tabIndex={0}
+                onClick={() => handleSuggestionClick(person)}
+                onKeyDown={(e) => {
+                  if (e.key === 'Enter') {
+                    handleSuggestionClick(person);
+                  }
```
volodymyr-soltys97, line 79: Move this logic from JSX to the helper function and use it here

src/App.tsx
```diff
@@ -1,54 +1,89 @@
...
+                }}
+              >
+                <p className={person === selectedPerson ? 'has-text-link' : ''}>
```
volodymyr-soltys97, line 82: Use the classnames library for add classes with condition

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code
Now people are shown all the time, and you need to show them only when the input is in the focus state
image

Remove it
image

Add the cursor pointer for list of the people
image

volodymyr-soltys97 (changes requested):
To improve:
The list does not disappear when the input is not in the focus state, this needs to be fixed
image

And now the list is not shown in focus state
image

SanyaBratashchuk (changes requested):
Looks like you have forgotten to implement 
image

SanyaBratashchuk (approved):

PR #454, opened 2024-01-19
add solution

nabokok:
[DEMO LINK](https://nabokok.github.io/react_autocomplete/)

### DIFF
--- a/src/components/DropdownInput/DropdownInput.tsx
+++ b/src/components/DropdownInput/DropdownInput.tsx
@@ -1,0 +1,45 @@
+import React, { useCallback } from 'react';
+import debounce from 'lodash.debounce';
+
+interface Props {
+  query: string;
+  setQuery: (query: string) => void
+  setAppliedQuery: (query: string) => void;
+  delay: number;
+  setIsListShown: (arg: boolean) => void;
+  onBlur: (e: React.FocusEvent<HTMLInputElement, Element>) => void;
+}
+
+export const DropdownInput: React.FC<Props> = ({
+  query,
+  setQuery,
+  setAppliedQuery,
+  delay,
+  setIsListShown,
+  onBlur,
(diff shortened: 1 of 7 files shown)


vSainiuk (approved):
Marvelous  💫

src/components/DropdownList/DropdownList.tsx
```diff
@@ -0,0 +1,51 @@
...
+
+  return (
+    <div className="dropdown-menu" role="menu">
+      <div className="dropdown-content">
+        {people.length
+          ? people.map((person) => (
+            <div
+              className="dropdown-item"
+              key={person.name}
+              onKeyDown={() => { }}
```
DanilWeda, line 32: Why do you need empty func in this place?

src/components/DropdownInput/DropdownInput.tsx
```diff
@@ -0,0 +1,45 @@
...
+
+  return (
+    <div className="dropdown-trigger">
+      <input
+        type="text"
+        placeholder="Enter a part of the name"
+        className="input"
+        value={query}
+        onChange={handleQueryChange}
+        onFocus={() => setIsListShown(true)}
```
DanilWeda, line 40: you can write a handler for it ( not critical )

DanilWeda (approved):
Great! All good! 
Approved! 
Many thanks. 

Also, you can fix height of dropdown: 
image

src/components/DropdownList/DropdownList.tsx
nabokok: I need a keyboard listener because of this eslint rule: "Visible, non-interactive elements with click handlers must have at least one keyboard listener". However I didn't implement any functionality when pressing the button, so I added an empty func. In short, I need it to avoid linter errors

PR #462, opened 2024-01-21
add task solution

dyarechuk:
[DEMO LINK](https://dyarechuk.github.io/react_autocomplete/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,54 +1,123 @@
-import React from 'react';
+import React, { useCallback, useState } from 'react';
 import './App.scss';
+import cn from 'classnames';
+import debounce from 'lodash.debounce';
 import { peopleFromServer } from './data/people';
+import { Person } from './types/Person';
 
 export const App: React.FC = () => {
-  const { name, born, died } = peopleFromServer[0];
+  const [selectedPerson, setSelectedPerson] = useState<Person | null>(null);
+  const [query, setQuery] = useState('');
+  const [appliedQuery, setAppliedQuery] = useState('');
+  const [visible, setVisible] = useState(false);
+
+  // eslint-disable-next-line
+  const applyQuery = useCallback(
+    debounce((value: string) => setAppliedQuery(value), 1000),
+    [],
+  );
+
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -1,54 +1,123 @@
...
+import debounce from 'lodash.debounce';
 import { peopleFromServer } from './data/people';
+import { Person } from './types/Person';
 
 export const App: React.FC = () => {
-  const { name, born, died } = peopleFromServer[0];
+  const [selectedPerson, setSelectedPerson] = useState<Person | null>(null);
+  const [query, setQuery] = useState('');
+  const [appliedQuery, setAppliedQuery] = useState('');
+  const [visible, setVisible] = useState(false);
```
Xilston, line 12: Use more understandable variables names. `visible` does not provide sense for what this variable.

src/App.tsx
```diff
@@ -1,54 +1,123 @@
...
+    setAppliedQuery('');
+    setQuery('');
+  };
+
+  const handleOnChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+    setQuery(event.target.value);
+    applyQuery(event.target.value);
+  };
+
+  const preparedPeople = (people: Person[], queryFilter: string) => {
```
Xilston, line 44: `preparedPeople` better to name it `preparePeople`, without ed at the end, because it's function.

Xilston (approved):
Gooj job! Nice solutions with reset button, I've enjoyed.

src/App.tsx
DanilWeda: Correct. Fix, please.

src/App.tsx
DanilWeda: Correct. Fix, please.

DanilWeda, line 44: you can move it outside component file

DanilWeda (changes requested):
Good job. 
Check some minor comments.
Many thanks!

olya-shyian (approved):
Nice!✨

PR #464, opened 2024-01-22
add task solution

vkolis:
[DEMO LINK](https://vk-workshop.github.io/react_autocomplete/)

### DIFF
--- a/src/AutoSuggest.tsx
+++ b/src/AutoSuggest.tsx
@@ -39,42 +39,42 @@ export const AutoSuggest: React.FC<Props> = ({
   const handleSuggestionClick = (selectedPerson: Person) => {
     setQuery(selectedPerson.name);
     onSelected(selectedPerson);
+    setAppliedQuery(selectedPerson.name);
   };
 
   return (
-    <>
-      <div className={cn('dropdown', { 'is-active': isInputFocused })}>
-        <div className="dropdown-trigger">
-          <input
-            type="text"
-            placeholder="Enter a part of the name"
-            className="input"
-            value={query}
-            onChange={handleQueryChange}
-            onBlur={() => setIsInputFocused(false)}
-            onFocus={() => setIsInputFocused(true)}
-          />
-        </div>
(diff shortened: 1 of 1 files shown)


src/AutoSuggest.tsx
```diff
@@ -0,0 +1,79 @@
...
+
+  const searchPerson = query.trim().toLowerCase();
+
+  const filteredPeople = useMemo(() => {
+    return peopleFromServer.filter(person => (
+      person.name.toLowerCase().includes(searchPerson)));
+  }, [peopleFromServer, appliedQuery]);
+
+  const applyQuery = useCallback(debounce((value: string) => {
+    setAppliedQuery(value);
```
l4st1m0za, line 29: Debounce function is not implemented correctly. You need to call the debounced function from within your event handler after creating it outside the scope of the event handler.

src/AutoSuggest.tsx
```diff
@@ -0,0 +1,79 @@
...
+  const [query, setQuery] = useState('');
+  const [appliedQuery, setAppliedQuery] = useState('');
+  const [isInputFocused, setIsInputFocused] = useState(false);
+
+  const searchPerson = query.trim().toLowerCase();
+
+  const filteredPeople = useMemo(() => {
+    return peopleFromServer.filter(person => (
+      person.name.toLowerCase().includes(searchPerson)));
+  }, [peopleFromServer, appliedQuery]);
```
l4st1m0za, line 26: The dependency array of useMemo should include 'searchPerson' instead of 'peopleFromServer' and 'appliedQuery'.

l4st1m0za (changes requested):
Good job, some small changes needed

etojeDenys (changes requested):
good job, but looks like you forgot to implement this option
image

vk-workshop:
- implemented hide suggestion on text change

lerastarynets (approved):
You're on the right track!

PR #467, opened 2024-01-23
solution

yepolotn1ak:
[DEMO LINK](https://yepolotn1ak.github.io/react_autocomplete/)

### DIFF
--- a/src/App.scss
+++ b/src/App.scss
@@ -1,3 +1,8 @@
 iframe {
   display: none;
 }
+
+.dropdown-trigger {
+  display: flex;
+  gap: 10px;
+}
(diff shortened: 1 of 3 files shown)


Andriy-Shymkiv (changes requested):
please fix this bug:
1. select person
2. press clear
3. open dropdown
expected: list of users
received: the last selected user
image

p.s.: sometimes I cannot select a user at all

loralevitska (changes requested):
Check my comments for improvements

- use [string.trim()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/trim) for search to avoid such bugs with multiple spaces
image
- fix warnings 
image

l4st1m0za (approved):
GJ!

PR #474, opened 2024-01-25
Task solution

Sireus:
[DEMO LINK](https://Sireus.github.io/react_autocomplete/)


src/App.tsx
```diff
@@ -1,56 +1,91 @@
-import React from 'react';
-import './App.scss';
-import { peopleFromServer } from './data/people';
+/* eslint-disable @typescript-eslint/quotes */
```
TarasHoliuk, line 1: just use single quotes instead of disabling lint rule

src/App.tsx
```diff
@@ -1,56 +1,91 @@
...
 
 export const App: React.FC = () => {
-  const { name, born, died } = peopleFromServer[0];
+  const [query, setQuery] = useState("");
+  const [appliedQuery, setAppliedQuery] = useState("");
+  const [selectedPerson, setSelectedPerson] = useState<Person | null>(null);
+  const [showMenu, setShowMenu] = useState(false);
+
+  // eslint-disable-next-line react-hooks/exhaustive-deps
+  const addQuery = useCallback(debounce(setAppliedQuery, 500), []);
```
TarasHoliuk, line 16:
```suggestion
  const applyQuery = useCallback(debounce(setAppliedQuery, 500), []);
```

src/App.tsx
```diff
@@ -1,56 +1,91 @@
...
+
+  const filteredPeople = useMemo(
+    () => peopleFromServer.filter(
+      person => person.name.toLowerCase().includes(appliedQuery.toLowerCase()),
+    ), [appliedQuery],
+  );
+
+  const handleQueryChange = (event: React.ChangeEvent<HTMLInputElement>) => {
```
TarasHoliuk, line 24: you should also change `showMenu` here to reach this requirement:
image

TarasHoliuk (changes requested):
GJ :+1: 
Let's implement 1 more thing (from task requirements)

src/App.tsx
Sireus: I'd love to, but for some reason my eslint automatically formats in double quotes. I haven't found a way to fix it...

src/App.tsx
Sireus: I seem to have taken this point into account and am changing it in the input. Isn't that how it should be?

image

src/App.tsx
```diff
@@ -1,56 +1,91 @@
...
+  const applyQuery = useCallback(debounce(setAppliedQuery, 500), []);
+
+  const filteredPeople = useMemo(
+    () => peopleFromServer.filter(
+      person => person.name.toLowerCase().includes(appliedQuery.toLowerCase()),
+    ), [appliedQuery],
+  );
+
+  const handleQueryChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+    if (appliedQuery.length === 0) {
```
IvanFesenko, line 25: means when user start typing text dropdown should hide and apear after delay
```suggestion
   setShowMenu(false)
    if (appliedQuery.length === 0) {
```

IvanFesenko (changes requested):

etojeDenys (changes requested):
image
means when user start typing text dropdown should hide and apear after delay

still not fixed

PR #480, opened 2024-01-27
add task solution

mykyta01:
- [DEMO LINK](https://mykyta01.github.io/react_autocomplete/)

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,15 +1,12 @@
 module.exports = {
-  extends: [
-    '@mate-academy/eslint-config-react-typescript',
-    'plugin:cypress/recommended',
-  ],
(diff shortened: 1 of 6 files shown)


volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code
When you click on a person from the list, nothing happens, this needs to be fixed
image

mykyta01:
> Good job 👍
> Let's improve your code
> When you click on a person from the list, nothing happens, this needs to be fixed
> image

Hi, 
I don't get what exactly should happen when I clicked on person?

volodymyr-soltys97 (changes requested):
Still not fixed from the previous review
After we click on a person in the input, we need to show this person in the input and above there should be data about him, now there is no
image

PR #484, opened 2024-01-29
autocompleted

AnoriFrell:
[DEMO LINK](https://AnoriFrell.github.io/react_autocomplete/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,99 +1,99 @@
-import React from 'react';
+import React, {
+  useCallback,
+  useEffect,
+  useMemo,
+  useState,
+} from 'react';
+import { debounce } from 'lodash';
 import './App.scss';
 import { peopleFromServer } from './data/people';
+import { Person } from './types/Person';
+import { PeopleList } from './components/People';
 
 export const App: React.FC = () => {
-  const { name, born, died } = peopleFromServer[0];
+  const [query, setQuery] = useState('');
+  const [appliedQuery, setAppliedQuery] = useState('');
+  const [isVisible, setIsVisible] = useState(false);
+  const [title, setTitle] = useState<string>('No selected person');
+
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -1,99 +1,99 @@
-import React from 'react';
+import React, {
+  useCallback,
+  useEffect,
+  useMemo,
+  useState,
+} from 'react';
+import { debounce } from 'lodash';
```
volodymyr-soltys97, line 7: Fix it
image

src/components/People.tsx
```diff
@@ -0,0 +1,51 @@
...
+      {people.length === 0 ? (
+        <div
+          className="
+        notification
+        is-danger
+        is-light
+        mt-3
+        is-align-self-flex-start
+      "
+          role="alert"
```
volodymyr-soltys97, line 26:
```suggestion
          className="
            notification
            is-danger
            is-light
            mt-3
            is-align-self-flex-start
          "
          role="alert"
```

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code
When you click on a person from the list, nothing happens, but you need to show the selected person in the input and the text at the top should change
image

anastasiiavorobiova (changes requested):
Great work! Unfortunately, I can't select/change a person. For some people, it doesn't work

AnoriFrell:
for some reason, the behavior of the page on the live server is different from that on the demo. Why so?

anastasiiavorobiova (changes requested):
Not fixed

etojeDenys (changes requested):
good job, but looks like you forgot to implement this option
image

volodymyr-soltys97 (approved):
Well done 🔥

PR #486, opened 2024-01-30
Develop

P-Nazar:
[DEMO LINK](https://P-Nazar.github.io/react_autocomplete/)

### DIFF
--- a/src/component/Autocomplete.tsx
+++ b/src/component/Autocomplete.tsx
@@ -22,10 +22,11 @@ export const Autocomplate: React.FC<Props> = ({
 }) => {
   const [query, setQuery] = useState('');
   const [appliedQuery, setAppliedQuery] = useState('');
+  const [delay] = useState(300);
   const [isInputFocused, setIsInputFocused] = useState(false);
 
   // eslint-disable-next-line react-hooks/exhaustive-deps
-  const applyQuery = useCallback(debounce(setAppliedQuery, 1000), []);
+  const applyQuery = useCallback(debounce(setAppliedQuery, delay), []);
 
   const handleQueryClick = (
     event: React.MouseEvent<HTMLAnchorElement, MouseEvent>,
@@ -50,8 +51,11 @@ export const Autocomplate: React.FC<Props> = ({
 
     setQuery(newQuery);
     applyQuery(newQuery);
-
     handleTitleChange(newQuery);
+    setIsInputFocused(false);
+    setTimeout(() => {
+      setIsInputFocused(true);
+    }, 1000);
   };
 
   const filteredPeople = useMemo(() => {


src/App.tsx
```diff
@@ -1,98 +1,62 @@
...
+    setQuery(event.target.value);
+    applyQuery(event.target.value);
+    setIsInputFocused(true);
+  };
+
+  const filteredPeople = useMemo(() => {
+    return peopleFromServer.filter(
+      person => person.name.toLowerCase().includes(appliedQuery.toLowerCase()),
+    );
+  }, [appliedQuery]);
```
VitaliyBondarenko1982, line 39: move all logic except selectedPerson state in Autocomplete component.
Also pay attention on this point from description:
`the delay should be customizable via props (default value is 300ms);`
it means that delay should come to Autocomplet component as not required prop with default value 300ms.
For better testing pass delay 1000ms.

src/App.tsx
```diff
@@ -1,98 +1,62 @@
...
+    person => person.name === selectedPerson,
+  );
+
+  const handleQueryClick = (
+    event: React.MouseEvent<HTMLAnchorElement, MouseEvent>,
+  ) => {
+    event.preventDefault();
+    setQuery(event.currentTarget.textContent || '');
+    setSelectedPerson(event.currentTarget.textContent || '');
+    setIsInputFocused(true);
```
VitaliyBondarenko1982, line 26:
```suggestion
    event.preventDefault();
    setQuery(event.currentTarget.textContent || '');
    setSelectedPerson(event.currentTarget.textContent || '');
```
no need to manage focus here. it should make onFocus and onBlur handlers.
better to pass person in function and set work with person or person.name instead event.currentTarget.textContent.

src/App.tsx
```diff
@@ -1,98 +1,62 @@
...
+  };
+
+  const handleQueryChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+    setQuery(event.target.value);
+    applyQuery(event.target.value);
+    setIsInputFocused(true);
```
VitaliyBondarenko1982, line 32:
```suggestion
```
do not manage focus here.

VitaliyBondarenko1982 (changes requested):
PLease check comments.
- After select person dropdown shouldn't be visible.
- after click on input (focus) and not start to type should be visible all suggestions

src/component/Autocomplete.tsx
```diff
@@ -0,0 +1,87 @@
...
+            data-cy="suggestions-list"
+          >
+            <div className="dropdown-content">
+              {filteredPeople.map(person => (
+                <a
+                  key={person.slug}
+                  href="."
+                  className="dropdown-item"
+                  data-cy="suggestion-item"
+                  onClick={onSelected}
```
VitaliyBondarenko1982, line 69:
```suggestion
                  onMouseDown={onSelected}
```
it will fix select, because onBlur work earlier then onClick

VitaliyBondarenko1982 (changes requested):

etojeDenys (changes requested):
looks like you forgot to implement this option
image



2. still not fixed (move all logic to Autocomplete.tsx)
image

anastasiiavorobiova (changes requested):
Great work! When the user changes a value in the input, reset the selected person

volodymyr-soltys97 (changes requested):
Almost done!
image

volodymyr-soltys97 (approved):
Good job 👍

PR #487, opened 2024-01-30
solution

InnaSh23:
[DEMO LINK](https://InnaSh23.github.io/react_autocomplete/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,57 +1,148 @@
-import React from 'react';
-import './App.scss';
+import React, { useState } from 'react';
+import 'bulma';
 import { peopleFromServer } from './data/people';
+import { Autocomplete } from './components/Autocomplete';
 
 export const App: React.FC = () => {
-  const { name, born, died } = peopleFromServer[0];
+  const [selectedPerson, setSelectedPerson] = useState<string | null>(null);
 
   return (
     <main className="section">
-      <h1 className="title">
-        {`${name} (${born} = ${died})`}
-      </h1>
+      {selectedPerson && (
+        peopleFromServer.map((person) => {
+          if (person.name === selectedPerson) {
+            return (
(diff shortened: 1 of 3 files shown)


anastasiiavorobiova (changes requested):
Great work! Consider hiding the list if the input is not focused

etojeDenys (changes requested):
image
you forgot to implement this option

src/App.tsx
```diff
@@ -1,57 +1,38 @@
...
+          if (foundPerson) {
+            return (
+              <h1 className="title" key={selectedPerson}>
+                {`${foundPerson.name} (${foundPerson.born} - ${foundPerson.died})`}
+              </h1>
+            );
+          }
+
+          return null;
+        })()
```
Andriy-Shymkiv, line 26: I don't like this anonymous function to be honest. Can you please make it just like a regular `jsx`?

Andriy-Shymkiv (changes requested):

src/components/Autocomplete.tsx
```diff
@@ -0,0 +1,91 @@
...
+  const handleSuggestion = (person: string) => {
+    setQuery(person);
+    setShowSuggestions(false);
+    onSelected(person);
+  };
+
+  const handleBlur = () => {
+    setTimeout(() => {
+      setShowSuggestions(false);
+    }, 200);
```
VitaliyBondarenko1982, line 50:
```suggestion
      setShowSuggestions(false);
```
if instead onClick set onMouseDown event - no need will be use setTimeout here)

VitaliyBondarenko1982 (approved):
Looks good.

src/components/Autocomplete.tsx
InnaSh23: Got it, thanks for the tip) I'll try to remember

PR #492, opened 2024-02-02
Solution

dmytro-popovych:
[DEMO LINK](https://dmytro-popovych.github.io/react_autocomplete/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,56 +1,91 @@
-import React from 'react';
 import './App.scss';
+import debounce from 'lodash.debounce';
+import classNames from 'classnames';
+import React, { useCallback, useMemo, useState } from 'react';
 import { peopleFromServer } from './data/people';
+import { Person } from './types/Person';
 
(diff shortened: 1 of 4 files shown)


src/App.tsx
```diff
@@ -1,56 +1,91 @@
...
+    setQuery(person.name);
+  };
 
   return (
     <main className="section">
       <h1 className="title">
-        {`${name} (${born} = ${died})`}
+        {selectedPerson
+          ? `${selectedPerson.name} (${selectedPerson.born} - ${selectedPerson.died})`
+          : 'No selected person'}
```
volodymyr-soltys97, line 45: Move this logic to the variable and use it here

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code
Need to clear the text above, after we enter the text in the input and then select the person and after that clear the input
image

sTorba24 (approved):

PR #512, opened 2024-02-20
create autocomplete

RomanSylman:
[DEMO LINK](https://RomanSylman.github.io/react_autocomplete/)

### DIFF
--- a/src/components/Autocomplete.tsx
+++ b/src/components/Autocomplete.tsx
@@ -1,0 +1,111 @@
+import React, {
+  useCallback, useMemo, useRef, useState,
+} from 'react';
+// eslint-disable-next-line
+import debounce from 'lodash.debounce';
+import { peopleFromServer } from '../data/people';
+import { Person } from '../types/Person';
+
+interface AutoCompleteProps {
+  onSelected: (person: Person | null) => void;
+}
+
+export const AutoComplete: React.FC<AutoCompleteProps> = ({ onSelected }) => {
+  const [inputValue, setInputValue] = useState('');
+  const [showAllPeople, setShowAllPeople] = useState(false);
+  const dropdownRef = useRef(null);
+
+  const applyQuery = useCallback(() => {
+    setShowAllPeople(true);
+  }, []);
+
+  const debouncedApplyQuery = debounce(applyQuery, 300);
+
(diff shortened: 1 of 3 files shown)


src/components/Autocomplete.tsx
```diff
@@ -0,0 +1,111 @@
+import React, {
+  useCallback, useMemo, useRef, useState,
+} from 'react';
+// eslint-disable-next-line
+import debounce from 'lodash.debounce';
+import { peopleFromServer } from '../data/people';
```
vadiimvooo, line 6: You don't need to directly import the data here. Just send it as props to Autocomplete component.

src/components/Autocomplete.tsx
```diff
@@ -0,0 +1,111 @@
...
+export const AutoComplete: React.FC<AutoCompleteProps> = ({ onSelected }) => {
+  const [inputValue, setInputValue] = useState('');
+  const [showAllPeople, setShowAllPeople] = useState(false);
+  const dropdownRef = useRef(null);
+
+  const applyQuery = useCallback(() => {
+    setShowAllPeople(true);
+  }, []);
+
+  const debouncedApplyQuery = debounce(applyQuery, 300);
```
vadiimvooo, line 22: Debounce time can be also passed as prop to the component.

vadiimvooo (changes requested):
Good job. A few things to fix.

anastasiiavorobiova (changes requested):
Great work! The list doesn't appear after blurring and focusing again. Could you consider fixing it? If you feel some of the comments are unclear, please ask for help in the chat or come to the Q&A session. We will deal with everything)

Esceype (changes requested):
Great job!
The last step is to add a clear button in the input, as shown in the [MUI example](https://mui.com/material-ui/react-autocomplete/#combo-box).
image

Esceype (changes requested):
GJ, please check if this works for you on the local machine, and redeploy if it works. because right now onFocus doesn't work on the demo, the list of people is visible only when I write something.

RomanSylman:
> GJ, please check if this works for you on the local machine, and redeploy if it works. because right now onFocus doesn't work on the demo, the list of people is visible only when I write something.

I guess it's GitHub delay

Esceype (approved):
Well done!

PR #548, opened 2024-03-14
Solution

wsadocha:
[DEMO LINK](https://wsadocha.github.io/react_autocomplete/)


src/components/Dropdown.tsx
```diff
@@ -0,0 +1,104 @@
...
+                  <p className="has-text-link">{person.name}</p>
+                </div>
+              ))}
+            </div>
+          </div>
+        ) : null}
+      </div>
+      {!filteredPeople.length && (
+        <div
+          className="
```
yevhenii-pyl, line 89: What happened to that className? :) 
className property should always be at one line (except of when you use classnames library with conditions for generating a className)

PR #560, opened 2024-03-18
ts autocompited

DenNice-cloud:
[DEMO LINK](https://DenNice-cloud.github.io/react_autocomplete/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,15 +1,54 @@
-import React from 'react';
+import React, { useState, useMemo } from 'react';
 import './App.scss';
 import { peopleFromServer } from './data/people';
+import debounce from 'lodash.debounce';
+import classNames from 'classnames';
+
+interface Person {
+  name: string;
+  born: number;
+  died: number;
+  sex: string;
+}
 
 export const App: React.FC = () => {
-  const { name, born, died } = peopleFromServer[0];
+  const [query, setQuery] = useState('');
+  const [selected, setSelected] = useState(false);
+  const [selectedPerson, setSelectedPerson] =
+    useState<string>('No selected person');
+
+  const applyQuery = () => {
+    debounce(() => setSelected(true), 300)();
+  };
+
+  const handleQueryChanged = (event: React.ChangeEvent<HTMLInputElement>) => {
+    setQuery(event.target.value);
+    setSelectedPerson('No selected person');
+  };
+
+  const filteredPeople = useMemo(() => {
+    const normalisedQuery = query.toLowerCase().trim();
+
+    return peopleFromServer.filter(person => {
+      const normalisedPerson = person.name.toLowerCase();
+
+      return normalisedPerson.includes(normalisedQuery);
+    });
+  }, [query]);
+
+  const handlePersonClick = (person: Person) => {
+    setQuery(person.name);
+    setSelected(false);
+    setSelectedPerson(() => {
+      return `${person.name} (${person.born} - ${person.died})`;
+    });
+  };
 
   return (
     <div className="container">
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -1,73 +1,108 @@
...
-
-              <div className="dropdown-item" data-cy="suggestion-item">
-                <p className="has-text-link">Pieter de Decker</p>
+            {selected && (
+              <div className="dropdown-content">
+                {filteredPeople.map(person => (
+                  <div
+                    className="dropdown-item"
+                    data-cy="suggestion-item"
+                    key={person.name}
```
anastasia1383, line 74: it's better to use slug for a key

src/App.tsx
```diff
@@ -1,73 +1,108 @@
...
               placeholder="Enter a part of the name"
+              value={query}
               className="input"
               data-cy="search-input"
+              onChange={handleQueryChanged}
+              onFocus={applyQuery}
             />
           </div>
 
           <div className="dropdown-menu" role="menu" data-cy="suggestions-list">
```
anastasia1383, line 67: it is said in the task that you should have delay as a prop, so probably you need to move dropdown logic to a custom component

src/App.tsx
```diff
@@ -1,73 +1,108 @@
-import React from 'react';
+import React, { useState, useMemo } from 'react';
 import './App.scss';
 import { peopleFromServer } from './data/people';
+import debounce from 'lodash.debounce';
+import classNames from 'classnames';
+
+interface Person {
```
anastasia1383, line 7: why do you need this interface? there is a Person.ts in types folder

anastasia1383 (changes requested):

DenNice-cloud:
> Good job! Not sure it was okay for this task to use the lodash lib, but great idea!

As i see, we can use it in code, I think thats no problem)

src/App.tsx
```diff
@@ -1,73 +1,19 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
-import { peopleFromServer } from './data/people';
+import { Dropdown } from './component/Dropdown';
 
 export const App: React.FC = () => {
-  const { name, born, died } = peopleFromServer[0];
+  const [selectedPerson, setSelectedPerson] =
+    useState<string>('No selected person');
```
BudnikOleksii, line 7:
```suggestion
  const [selectedPerson, setSelectedPerson] =
    useState<Person | null>(null);
```
You should store the person instead of string

src/component/Dropdown.tsx
```diff
@@ -0,0 +1,96 @@
...
+  delay: number;
+};
+
+export const Dropdown: React.FC<Props> = ({ onSelected, delay = 300 }) => {
+  const [query, setQuery] = useState('');
+  const [selected, setSelected] = useState(false);
+
+  const applyQuery = () => {
+    debounce(() => setSelected(true), delay)();
+  };
```
BudnikOleksii, line 19: - suggestions should appear after some delay in typing (debounce);
image

You should set `appliedQuery` and filter people depending on it. Now it isn't working.
`selected` state is redundant and doesn't work.

BudnikOleksii (changes requested):
Good start 👍 
But you don't filter people with delay, and store string instead of person. Let's fix this issues

src/App.tsx
BudnikOleksii: > You should store the person instead of string

Not fixed

BudnikOleksii (changes requested):
Great, but you still store string instead of person

StasSokolov1 (approved):
Looks good to me🔥

PR #563, opened 2024-03-20
solution

Patryk-Buczkowski:
[DEMO LINK](https://Patryk-Buczkowski.github.io/react_autocomplete/)

### DIFF
--- a/src/Dropdown.tsx
+++ b/src/Dropdown.tsx
@@ -1,0 +1,100 @@
+import React, { useMemo, useState } from 'react';
+import { Person } from './types/Person';
+import cn from 'classnames';
+import debounce from 'lodash.debounce';
+
+type Props = {
+  delay: number;
+  people: Person[];
+  onSelected: (person: Person | null) => void;
+};
+
+export const Dropdown: React.FC<Props> = ({ delay, people, onSelected }) => {
+  const [inputValue, setInputValue] = useState('');
+  const [visible, setVisible] = useState(false);
+
+  const debounceName = useMemo(
+    () => debounce(() => setVisible(true), delay),
+    [delay],
+  );
+
+  const filteredPeople = useMemo(() => {
+    return people.filter(person => {
+      const fixedText = inputValue.toLocaleLowerCase().trim();
+
+      return person.name.toLowerCase().includes(fixedText);
+    });
+  }, [people, inputValue]);
+
+  const handleSelectPerson = (person: Person) => {
+    onSelected(person);
+    setInputValue(person.name);
+    setVisible(false);
+  };
+
+  const handleInputValue = (event: React.ChangeEvent<HTMLInputElement>) => {
+    debounceName();
+    setInputValue(event.target.value);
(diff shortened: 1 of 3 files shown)


src/Dropdown.tsx
```diff
@@ -0,0 +1,100 @@
...
+export const Dropdown: React.FC<Props> = ({ delay, people, onSelected }) => {
+  const [inputValue, setInputValue] = useState('');
+  const [visible, setVisible] = useState(false);
+
+  const debounceName = useMemo(
+    () => debounce(() => setVisible(true), delay),
+    [delay],
+  );
+
+  const filteredPeople = useMemo(() => {
```
yevhenii-pyl, line 21: Not sure that you really want to use useMemo here, caching here may only decrease the performance.

src/Dropdown.tsx
```diff
@@ -0,0 +1,100 @@
...
+                      {person.name}
+                    </p>
+                  </div>
+                ))}
+            </div>
+          )}
+        </div>
+      </div>
+
+      {filteredPeople.length === 0 && (
```
yevhenii-pyl, line 83: Well, if you still prefer this way to check :) 

But I suggest to stick to `!!filteredPeople.length `

yevhenii-pyl (changes requested):
Good job! 
But we still need to update your repo to get these tests going, could you do it? :)

This branch is [1 commit ahead of](https://github.com/Patryk-Buczkowski/react_autocomplete/compare/mate-academy%3Areact_autocomplete%3Amaster...develop), [1 commit behind](https://github.com/Patryk-Buczkowski/react_autocomplete/compare/develop...mate-academy%3Areact_autocomplete%3Amaster) mate-academy/react_autocomplete:master.

src/Dropdown.tsx
Patryk-Buczkowski: seems that `!filteredPeople.length` works better ;)

src/Dropdown.tsx
Patryk-Buczkowski: without memo it do not work as I expect ;)

src/Dropdown.tsx
yevhenii-pyl: Yep, was looking on another condition you have :) 
Basically, there are two checks in your code: 
1. To show items when there are filteredPeople
2. Show a message when there are no filteredPeople

For the first one you have `filteredPeople .length > 0` => we can shorten it to `!!filteredPeople.length` => which basically means 'give me a boolean analog of this value' , so whenever it's > 0 (!!0 === false) we will display items 

And for the second one we just want to check that there is no value(length === 0), in this case it's just 
`! filteredPeople.length` => which means 'if there is NO value for the array length', this will evaluate to true only when filteredPeople === []

yevhenii-pyl (approved):
Nice one here! 
Now you are a Git-pro with these constant fork sync and merging 😄

Patryk-Buczkowski:
thanks ;)

PR #569, opened 2024-03-21
solution

iamgranovskaya:
[DEMO](https://iamgranovskaya.github.io/react_autocomplete/)


etojeDenys (approved):
better to add the logic to hide the dropdown on the input unfocus

iamgranovskaya:
> better to add the logic to hide the dropdown on the input unfocus 

How can I do this?
when I add onBlur on input el onBlur event triggers before onClick on dropdown item , how to prevent it?

PR #576, opened 2024-03-26
Solution

Aksahir:
[DEMO LINK](https://Aksahir.github.io/react_autocomplete/)


l4st1m0za (changes requested):
Good job, but you some wrong logic:

1) Your table of names should be closed if input is not on focus. 
2) When I have selected the name, table should not show all the available names, but those which include text from the input. In case of full name typed in the input, there should be only one name. 
3) On the first click on the input, your tables shows up for a sec and then dissapeares, and you are not able to use autocomlite, until you click second time

etojeDenys (approved):
lgtm

PR #583, opened 2024-04-02
React autocomplete

Denys209818:
[DEMO LINK](https://denys209818.github.io/react_autocomplete)


l4st1m0za (changes requested):
Everyhting seems to be good, the only thing to fix is (screenshot). When I enter the person which is not in the list, it shuold not show the list

l4st1m0za (approved):
Good job!

PR #596, opened 2024-04-07
cope

true-binar-White-Rabbit:
[DEMO LINK](https://true-binar-White-Rabbit.github.io/react_autocomplete/) 

Hi, everyone!
Please kindly request! 
Before talking about the requirement to pass automatic tests, check out the proposed variant. 
Review the suggested code and the approach to solving the original task. 
Suggest code optimization options and formulate the shortcomings of the proposed decision.
Ty!


src/App.tsx
```diff
@@ -1,16 +1,60 @@
 import React from 'react';
 import './App.scss';
 import { peopleFromServer } from './data/people';
+import { Person } from './types/Person';
 
 export const App: React.FC = () => {
-  const { name, born, died } = peopleFromServer[0];
+  const [query, setQuery] = React.useState('');
+  const [isDropedDown, setIsDropedDown] = React.useState(false);
+  const [selectedPerson, setSelectedPerson] = React.useState<Person | null>(
```
anastasiiavorobiova, line 9:
```suggestion
  const [query, setQuery] = useState('');
  const [isDropedDown, setIsDropedDown] = useState(false);
  const [selectedPerson, setSelectedPerson] = useState<Person | null>(
```

src/App.tsx
```diff
@@ -1,16 +1,60 @@
 import React from 'react';
```
anastasiiavorobiova, line 1:
```suggestion
import React, { useState } from 'react';
```

src/App.tsx
```diff
@@ -19,55 +63,76 @@ export const App: React.FC = () => {
...
+              is-danger
+              is-light
+              mt-3
+              is-align-self-flex-start"
+            role="alert"
+            data-cy="no-suggestions-message"
+          >
+            <p className="has-text-danger">No matching suggestions</p>
+          </div>
+        ) : selectedPerson && getFilterPeople().length > 0 ? (
```
anastasiiavorobiova, line 122: Avoid using nested ternary

anastasiiavorobiova (changes requested):
Great start! Keep doing! Feel free to ask for help in the chat to pass tests

src/App.tsx
true-binar-White-Rabbit: ty, but this is a matter of taste, not rationality.

src/App.tsx
true-binar-White-Rabbit: sorry, but this is not nested ternary, i use the only one ternary expression that compares two conditions via the operator ?: to determine which JSX block to render

anastasiiavorobiova (changes requested):
If you feel some of the comments are unclear, please ask for help in the chat or come to the Q&A session. We will deal with everything)

src/App.tsx
maxim2310: yes it's actually called nested ternary, try to rewrite without this antipattern

src/App.tsx
maxim2310: But would be better stick to react doc

maxim2310 (changes requested):
Good work, but lets make some improvement 
try to figure out with tests 
image

true-binar-White-Rabbit:
> Good work, but lets make some improvement try to figure out with tests image

is my title not updated after a person is selected? or when you delete the name, the title "No selected person" does not appear?

PR #608, opened 2024-04-16
Solition

Galers:
[DEMO LINK](https://Galers.github.io/react_autocomplete/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,73 +1,39 @@
-import React from 'react';
+import React, { useCallback, useState } from 'react';
 import './App.scss';
 import { peopleFromServer } from './data/people';
+import { Autocomplete } from './components/Autocomplete';
+import { Person } from './types/Person';
+
+const START_PHRASE = 'No selected person';
 
 export const App: React.FC = () => {
-  const { name, born, died } = peopleFromServer[0];
+  const [selectedPerson, setSelectedPerson] = useState<Person | null>(null);
+
+  const handleSelected = useCallback((person: Person) => {
+    setSelectedPerson(person);
+  }, []);
+
+  const handleInputChange = (inputValue: string) => {
+    if (inputValue.trim() === '') {
+      setSelectedPerson(null);
+    } else {
+      setSelectedPerson(null);
+    }
+  };
+
+  const { name, born, died } = selectedPerson || {};
 
   return (
     <div className="container">
       <main className="section is-flex is-flex-direction-column">
         <h1 className="title" data-cy="title">
-          {`${name} (${born} - ${died})`}
+          {selectedPerson ? `${name} (${born} - ${died})` : START_PHRASE}
         </h1>
-
-        <div className="dropdown is-active">
-          <div className="dropdown-trigger">
-            <input
-              type="text"
-              placeholder="Enter a part of the name"
-              className="input"
-              data-cy="search-input"
-            />
-          </div>
-
-          <div className="dropdown-menu" role="menu" data-cy="suggestions-list">
-            <div className="dropdown-content">
-              <div className="dropdown-item" data-cy="suggestion-item">
-                <p className="has-text-link">Pieter Haverbeke</p>
-              </div>
-
-              <div className="dropdown-item" data-cy="suggestion-item">
-                <p className="has-text-link">Pieter Bernard Haverbeke</p>
-              </div>
-
-              <div className="dropdown-item" data-cy="suggestion-item">
-                <p className="has-text-link">Pieter Antone Haverbeke</p>
-              </div>
-
-              <div className="dropdown-item" data-cy="suggestion-item">
-                <p className="has-text-danger">Elisabeth Haverbeke</p>
-              </div>
-
-              <div className="dropdown-item" data-cy="suggestion-item">
-                <p className="has-text-link">Pieter de Decker</p>
-              </div>
-
-              <div className="dropdown-item" data-cy="suggestion-item">
-                <p className="has-text-danger">Petronella de Decker</p>
-              </div>
-
-              <div className="dropdown-item" data-cy="suggestion-item">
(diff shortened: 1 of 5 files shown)


src/App.tsx
```diff
@@ -1,73 +1,39 @@
...
+  const handleSelected = useCallback((person: Person) => {
+    setSelectedPerson(person);
+  }, []);
+
+  const handleInputChange = (inputValue: string) => {
+    if (inputValue.trim() === '') {
+      setSelectedPerson(null);
+    } else {
+      setSelectedPerson(null);
+    }
```
sTorba24, line 21: what is the difference in actions?

src/components/Autocomplete.tsx
```diff
@@ -0,0 +1,149 @@
...
+    };
+
+    const handleClickOutside = () => {
+      setShowSuggestions(false);
+    };
+
+    const ref = useOutsideClick(handleClickOutside);
+
+    useEffect(() => {
+      if (appliedInputText === '') {
```
sTorba24, line 49:
```suggestion
      if (!appliedInputText) {
```

src/components/Autocomplete.tsx
```diff
@@ -0,0 +1,149 @@
...
+    };
+
+    const handleItemClick = (person: Person) => {
+      setInputText(person.name);
+      onSelected(person);
+      setShowSuggestions(false);
+    };
+
+    const handleInputFocus = () => {
+      if (inputText === '') {
```
sTorba24, line 80:
```suggestion
      if (!inputText) {
```

src/components/Autocomplete.tsx
```diff
@@ -0,0 +1,149 @@
...
+                    >
+                      {person.name}
+                    </p>
+                  </div>
+                ))}
+              </div>
+            )}
+          </div>
+        </div>
+        {filteredPeople.length === 0 && (
```
sTorba24, line 131:
```suggestion
        {!filteredPeople.length && (
```

sTorba24 (changes requested):
Almost good! Lets improve it

Show all suggestions when option selected and input focused
image


also add some styles to selected option

In example from readme file you can find how it works

src/App.tsx
```diff
@@ -1,73 +1,37 @@
...
+  const [selectedPerson, setSelectedPerson] = useState<Person | null>(null);
+
+  const handleSelected = useCallback((person: Person) => {
+    setSelectedPerson(person);
+  }, []);
+
+  const handleInputChange = (inputValue: string) => {
+    if (inputValue.trim() || inputValue.trim() === '') {
+      setSelectedPerson(null);
+    }
```
sTorba24, line 19: if inputValue.trim() === '' then inputValue.trim() can't give you truthy value, this if doesn't make any sense

src/components/Autocomplete.tsx
sTorba24: image
still not updated

sTorba24 (changes requested):
Fix my previous comment about showing suggestions when option selected

src/App.tsx
Galers: But if delete the name immediately, state of the title doesn't change. -_-

src/App.tsx
```diff
@@ -1,73 +1,37 @@
...
+  };
+
+  const { name, born, died } = selectedPerson || {};
 
   return (
     <div className="container">
       <main className="section is-flex is-flex-direction-column">
         <h1 className="title" data-cy="title">
-          {`${name} (${born} - ${died})`}
+          {selectedPerson ? `${name} (${born} - ${died})` : START_PHRASE}
```
volodymyr-soltys97, line 28: Move this logic to the helper variable and use it here

src/components/Autocomplete.tsx
```diff
@@ -0,0 +1,160 @@
...
+        </div>
+        {!filteredPeople.length && (
+          <div
+            className="
+            notification
+            is-danger
+            is-light
+            mt-3
+            is-align-self-flex-start
+          "
```
volodymyr-soltys97, line 150:
```suggestion
            className="
              notification
              is-danger
              is-light
              mt-3
              is-align-self-flex-start
            "
```

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code
You need to fix it 
image
image
You need to show all people after delete the person

Remove this block when we don't have people
image
image

src/components/Autocomplete.tsx
```diff
@@ -0,0 +1,172 @@
...
+        setShowSuggestions(filtered.length >= 0);
+      }
+    }, [appliedInputText, people]);
+
+    const handleInputChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+      setInputText(event.target.value);
+      applyInputText(event.target.value);
+
+      onInputChange(event.target.value);
+    };
```
sTorba24, line 81:
```suggestion
    const handleInputChange = (event: React.ChangeEvent<HTMLInputElement>) => {
      const { value }= event.target
      setInputText(value);
      applyInputText(value);

      onInputChange(value);
    };
```

sTorba24 (approved):
Now looks really cool! Approved :)

src/components/Autocomplete.tsx
```diff
@@ -0,0 +1,172 @@
...
+      if (!appliedInputText) {
+        setFilteredPeople(people);
+        setShowSuggestions(true);
+      } else {
+        const filtered = people.filter(person =>
+          person.name.toLowerCase().includes(appliedInputText.toLowerCase()),
+        );
+
+        setFilteredPeople(filtered);
+        setShowSuggestions(filtered.length >= 0);
```
sTorba24, line 72:
```
       filtered.length >= 0 // is always true. Length always giving number in this range
```

PR #617, opened 2024-04-24
Develop

DaniilBarilotti:
https://daniilbarilotti.github.io/react_autocomplete/


anastasiiavorobiova (changes requested):
Great work! Consider hiding the suggestions list while the user is typing

PR #619, opened 2024-04-25
solution

ArtemZakhar:
Hi.

All tests are passed.
But I can't deploy because I can turn off the below mistake: 
`Property 'name' does not exist on type '{} | Person'.`

`{Object.hasOwn(selectedPerson, 'name')
            ? `${selectedPerson.name} (${selectedPerson.born} - ${selectedPerson.died})`
            : `No selected person`}`

I tried: Object.keys, selectedPerson === undefined, in selectedPerson, QNA session, but nothing helped(

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,15 +1,43 @@
-import React from 'react';
-import './App.scss';
+import React, { useCallback, useState } from 'react';
+import debounce from 'lodash.debounce';
 import { peopleFromServer } from './data/people';
 
+import { Person } from './types/Person';
+
+import './App.scss';
+
 export const App: React.FC = () => {
-  const { name, born, died } = peopleFromServer[0];
+  const [activeDropDown, setActiveDropDown] = useState(false);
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -1,15 +1,43 @@
...
+import { Person } from './types/Person';
+
+import './App.scss';
+
 export const App: React.FC = () => {
-  const { name, born, died } = peopleFromServer[0];
+  const [activeDropDown, setActiveDropDown] = useState(false);
+  const [searchInput, setSearchInput] = useState('');
+  const [querry, setQuerry] = useState('');
+  const [selectedPerson, setSelectedPerson] = useState<Person | {}>({});
```
polosanya, line 13: Better set null as an initial value and just check if the value is not null

polosanya (approved):
Good job! I'm approving it, but please try to handle the problem

src/App.tsx
ArtemZakhar: Thank you!)

PR #621, opened 2024-04-27
Solution

robsza1122:
https://robsza1122.github.io/react_autocomplete/


BogdanMaliuta (changes requested):
Your demo doesn`t work, fix it please

robsza1122:
It should work.

src/components/Dropdown.tsx
```diff
@@ -0,0 +1,92 @@
...
+export const Dropdown: React.FC<DropdownType> = ({
+  people,
+  onSelected,
+  delay,
+}: DropdownType) => {
+  const [inputSearch, setInputSearch] = useState('');
+  const [searchedText, setSearchedText] = useState('');
+  const [isActive, setIsActive] = useState(false);
+
+  const filteredPeople = useMemo(() => {
```
yevhenii-pyl, line 17: No need in memoizaition here.

src/components/Dropdown.tsx
```diff
@@ -0,0 +1,92 @@
...
+                <p className="has-text-link">{option.name}</p>
+              </div>
+            ))}
+          </div>
+        </div>
+      </div>
+
+      {filteredPeople.length === 0 && (
+        <div
+          className="
```
yevhenii-pyl, line 77: props can (and should be) in column, when you have >~3 props.
however, prop value should always be in a row

` className="notification is-danger is-light mt-3 is-align-self-flex-start"`

yevhenii-pyl (changes requested):
Good job, however, few things to improve:

1. Please, pay attention to your imports. It should be structured as:

-- third-party libs

-- your libs

```
import react ...

import Component from 'MyComponents/Component'
```

2. Revise your approach with memoization, check this article for the detailed explanation: https://kentcdodds.com/blog/usememo-and-usecallback

PR #623, opened 2024-04-29
add_solution

metrakonon:
[DEMO](https://metrakonon.github.io/react_autocomplete/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,73 +1,28 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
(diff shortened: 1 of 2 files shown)


ArtemZakhar (approved):
Very interesting solution)

DanilWeda (changes requested):
Great job. 
In case when you choose input clearing, but in the result dropdown we have the previous sorted value. Please check it. Also, you can make the dropdown smaller on height ( it is too large and I have an overflow on the body) 
Thanks)

mrkovaliv (approved):
👍 👍 👍

PR #624, opened 2024-04-29
add task solution

verandetta:
[DEMO LINK](https://verandetta.github.io/react_autocomplete/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,73 +1,111 @@
-import React from 'react';
+import React, {
+  useCallback,
+  useEffect,
+  useMemo,
+  useRef,
+  useState,
+} from 'react';
+import debounce from 'lodash.debounce';
 import './App.scss';
 import { peopleFromServer } from './data/people';
+import { Person } from './types/Person';
+import classNames from 'classnames';
 
 export const App: React.FC = () => {
-  const { name, born, died } = peopleFromServer[0];
+  const titleField = useRef<HTMLInputElement>(null);
+
+  useEffect(() => {
+    if (titleField.current) {
+      titleField.current.focus();
+    }
+  }, []);
+
+  const [query, setQuery] = useState('');
+  const [appliedQuery, setAppliedQuery] = useState('');
+
+  const [selectedPerson, setSelectedPerson] = useState<Person | null>(null);
+
+  const filteredPersonList = useMemo(() => {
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -1,73 +1,111 @@
...
+  const [selectedPerson, setSelectedPerson] = useState<Person | null>(null);
+
+  const filteredPersonList = useMemo(() => {
+    return peopleFromServer.filter(person =>
+      person.name.toLowerCase().includes(appliedQuery.toLowerCase()),
+    );
+  }, [appliedQuery]);
+
+  // eslint-disable-next-line react-hooks/exhaustive-deps
+  const applyQuery = useCallback(debounce(setAppliedQuery, 2000), []);
```
etojeDenys, line 35:
```suggestion
  const applyQuery = useCallback(debounce(setAppliedQuery, 300), []);
```
image

etojeDenys (changes requested):
1. image
move this message inside the dropdown

2. close the dropdown list if the input unFocus or the user clicks on the person
image

3. move all dropdown logic to the AutoComplete component
image

src/App.tsx
```diff
@@ -1,73 +1,25 @@
...
+  const [selectedPerson, setSelectedPerson] = useState<Person | null>(null);
 
   return (
     <div className="container">
       <main className="section is-flex is-flex-direction-column">
         <h1 className="title" data-cy="title">
-          {`${name} (${born} - ${died})`}
+          {selectedPerson
+            ? `${selectedPerson.name} (${selectedPerson.born} - ${selectedPerson.died})`
+            : 'No selected person'}
```
volodymyr-soltys97, line 16: You can move this logic to the helper variable and use it here

volodymyr-soltys97 (changes requested):
Good job!
Let's improve your code.
You need to hide the dropdown menu when input isn't focused
image

The error message should be a separate block below, and the input should stay in the same place and style
image

vadiimvooo (changes requested):
Seems like there is one bug you have to fix. Please feel free to ask for help in fe_chat.
https://www.loom.com/share/9dca44fd26f94fd187b83d4eec2f1cb9

verandetta:
I`ve redeployed my task, so it should show correct version

volodymyr-soltys97 (approved):
Great work 🔥

PR #625, opened 2024-04-30
task solution

BoGoBooM:
[SOLUTION LINK](https://bogoboom.github.io/react_autocomplete/)


src/App.tsx
```diff
@@ -1,73 +1,25 @@
...
+  };
 
   return (
     <div className="container">
       <main className="section is-flex is-flex-direction-column">
         <h1 className="title" data-cy="title">
-          {`${name} (${born} - ${died})`}
+          {selectedPerson
+            ? `${selectedPerson.name} (${selectedPerson.born} - ${selectedPerson.died})`
+            : 'No selected person'}
```
volodymyr-soltys97, line 19: Move this logic from jsx to the helper variable and use it here

src/components/dropdown.tsx
```diff
@@ -0,0 +1,130 @@
...
+          <input
+            type="text"
+            ref={inputElement}
+            placeholder="Enter a part of the name"
+            className="input"
+            data-cy="search-input"
+            onChange={onInputChange}
+            onFocus={() => {
+              setIsActive(true);
+            }}
```
volodymyr-soltys97, line 73: Fix it everywhere
```suggestion
            onFocus={() => setIsActive(true)}
```

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's make your code better

PR #629, opened 2024-05-02
add solution

yanamomot:
[DEMO LINK](https://yanamomot.github.io/react_autocomplete/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,15 +1,38 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
 import { peopleFromServer } from './data/people';
+import { Person } from './types/Person';
 
 export const App: React.FC = () => {
-  const { name, born, died } = peopleFromServer[0];
+  const [query] = useState('');
+  const [selectUser, setSelectUser] = useState<Person | null>(null);
+  const [searchUser, setSearchUser] = useState('');
+  const [dropDown, setDropDown] = useState(false);
+
+  const handleQuery = (event: React.ChangeEvent<HTMLInputElement>) => {
+    setSelectUser(null);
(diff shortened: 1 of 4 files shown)


src/App.tsx
```diff
@@ -1,15 +1,44 @@
-import React from 'react';
+/* eslint-disable */
```
mrkovaliv, line 1: this isn't good practice disable eslint for whole file

mrkovaliv: use only where you need this but try to avoid disabling eslint

src/App.tsx
```diff
@@ -19,55 +48,55 @@ export const App: React.FC = () => {
...
-                <p className="has-text-link">Pieter de Decker</p>
-              </div>
-
-              <div className="dropdown-item" data-cy="suggestion-item">
-                <p className="has-text-danger">Petronella de Decker</p>
-              </div>
-
-              <div className="dropdown-item" data-cy="suggestion-item">
-                <p className="has-text-danger">Elisabeth Hercke</p>
+          {dropDown && (
```
mrkovaliv, line 57: dropDown => isDropDownVisible

src/App.tsx
```diff
@@ -19,55 +48,55 @@ export const App: React.FC = () => {
...
+                ))}
               </div>
             </div>
-          </div>
+          )}
         </div>
 
-        <div
-          className="
+        {filter.length === 0 && (
```
mrkovaliv, line 85: `filter.length && (`

src/App.tsx
mrkovaliv: also move on dropdown to separate component)))

mrkovaliv (approved):
Good job 👍 

Fix please all comments)))

PR #637, opened 2024-05-07
adding task solution

eduvinagre:
[DEMO LINK](https://eduvinagre.github.io/react_autocomplete/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,73 +1,109 @@
-import React from 'react';
+import { useState, useRef } from 'react';
 import './App.scss';
 import { peopleFromServer } from './data/people';
+import { Person } from './types/Person';
+import debounce from 'lodash.debounce';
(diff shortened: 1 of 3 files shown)


jv-aquino (approved):
Great job Edu! Everything is perfect as the project demanded, below I will just give some a11y tips / best practices that you may want to follow for real world projects - it's a good thing to keep in mind

The tip is: When we want to create clickable things, we generally want to inform the screen reader that it's a button, otherwise people will not be able to access it with tab (as it's occurring in the page)
Also in general it's cool to use cursor pointer for buttons or clickable stuff 
image

eduvinagre:
@jv-aquino Nice heads up, Mate! Most of the time I don't look for improvements like these because I'm too afraid to fail the tests, but I'll try to look more into this things and push my code to the limit.
Cheers!

PR #646, opened 2024-05-09
add task solution

Esther-Guimaraes:
[DEMO LINK](https://Esther-Guimaraes.github.io/react_autocomplete/)


src/App.tsx
```diff
@@ -1,20 +1,58 @@
...
+    if (inputName.trim() === '') {
+      setPersonList(peopleFromServer);
+    } else {
+      setPersonList(
+        peopleFromServer.filter(person =>
+          person.name.toLowerCase().includes(appliedInput.toLowerCase()),
+        ),
+      );
+    }
+  }, [appliedInput]);
```
EdPirro, line 29: Missing `inputName` in `React.useEffect`'s the dependency array

src/App.tsx
```diff
@@ -1,20 +1,58 @@
...
-  const { name, born, died } = peopleFromServer[0];
+  const [selectedName, setSelectedName] =
+    useState<string>('No selected person');
+  const [personList, setPersonList] = useState(peopleFromServer);
+  const [inputName, setInputName] = useState('');
+  const [appliedInput, setAppliedInput] = useState('');
+  const [isInputFocus, setIsInputFocus] = useState(true);
+
+  const applyInput = useCallback((value: string) => {
+    debounce(() => setAppliedInput(value), 300)();
```
EdPirro, line 16: Delay should be customizable through a prop [Reference](https://github.com/mate-academy/react_autocomplete#:~:text=suggestions%20should%20appear,value%20is%20300ms\)%3B)

src/App.tsx
```diff
@@ -1,20 +1,58 @@
...
-          {`${name} (${born} - ${died})`}
+          {selectedName}
         </h1>
 
-        <div className="dropdown is-active">
+        <div className={`dropdown ${isInputFocus ? 'is-active' : ''} `}>
           <div className="dropdown-trigger">
             <input
+              value={inputName}
+              onClick={() => setIsInputFocus(true)}
```
EdPirro, line 54: You should use [`onFocus`](https://developer.mozilla.org/en-US/docs/Web/API/Element/focus_event) and [`onBlur`](https://developer.mozilla.org/en-US/docs/Web/API/Element/blur_event) to manage input focus

src/App.tsx
```diff
@@ -24,50 +62,34 @@ export const App: React.FC = () => {
...
+              is-light
+              mt-3
+              is-align-self-flex-start
+            "
+            role="alert"
+            data-cy="no-suggestions-message"
+          >
+            <p className="has-text-danger">No matching suggestions</p>
+          </div>
+        ) : null}
```
EdPirro, line 92: This is being rendered outside the dropdown, which causes a visual glitch

image

EdPirro (changes requested):
Good effort, @Esther-Guimaraes! You're almost there 👏

Just a few touch-ups and it should be good to go! Let me know if you have any questions!

src/App.tsx
Esther-Guimaraes: Hello, Edu! I had tried this initially and noticed that there was a conflict between OnBlur and OnClick. Anyway, I found a way to make the two work together using setTimeout. I believe that with UseEffect it would also work. Thanks!

Esther-Guimaraes:
As mentioned above, I managed to set onBlur and onFocus using setTimeout, the tests passed on my PC, but they didn't run at all on the git server.  I went back to the shape I was in, because that's the only way it worked.

src/App.tsx
```diff
@@ -1,20 +1,62 @@
...
+  const [selectedName, setSelectedName] =
+    useState<string>('No selected person');
+  const [personList, setPersonList] = useState(peopleFromServer);
+  const [inputName, setInputName] = useState('');
+  const [appliedInput, setAppliedInput] = useState('');
+  const [isInputFocus, setIsInputFocus] = useState(true);
+
+  const applyInput = useCallback((value: string) => {
+    debounce(() => setAppliedInput(value), debounceDelay)();
+  }, []);
```
jv-aquino, line 21: missing dependency here

PR #648, opened 2024-05-13
Task autocomplete

HigorPedroso:
[DEMO LINK](https://HigorPedroso.github.io/react_autocomplete/)

### DIFF
--- a/src/components/Autocomplete/Autocomplete.tsx
+++ b/src/components/Autocomplete/Autocomplete.tsx
@@ -1,9 +1,9 @@
-import React from "react"
-import { memo, useCallback, useEffect, useMemo, useState } from 'react';
-import { Person } from "../../types/Person";
-import { peopleFromServer } from "../../data/people";
-import debounce from "lodash.debounce";
-import classNames from "classnames";
+import React from 'react';
+import { memo, useEffect, useMemo, useState } from 'react';
+import { Person } from '../../types/Person';
+import { peopleFromServer } from '../../data/people';
+import debounce from 'lodash.debounce';
+import classNames from 'classnames';
 
 interface Props {
   delay?: number;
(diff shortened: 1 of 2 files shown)


src/components/Autocomplete/Autocomplete.tsx
```diff
@@ -0,0 +1,127 @@
...
+    }, [filteredPeople]);
+
+    const applyQuery = useCallback(
+      debounce((value: string) => {
+        if (value !== query) {
+          setAppliedQuery(value);
+        }
+      }, delay),
+      [query],
+    );
```
EdPirro, line 53: `useCallback` is being unable to compute the debounce return's dependencies, `useCallback` expects an inline function as parameter.
Change this to a `useMemo` with an inline function:
```js
const applyQuery = useMemo(() => (
  debounce(
    (value: string) => {
      if (value !== query) {
        setAppliedQuery(value);
      }
    }, 
    delay
  )),
  [query]
);
```

src/components/Autocomplete/Autocomplete.tsx
```diff
@@ -0,0 +1,127 @@
...
+          onWarning(true);
+          setShowDrop(true);
+        } else {
+          onWarning(false);
+          setShowDrop(false);
+        }
+      } else {
+        setIsFirstRender(false);
+      }
+    }, [filteredPeople]);
```
EdPirro, line 44: `isFirstRender` and `onWarning` are missing in the dependency array

EdPirro (changes requested):
Good job, @HigorPedroso! I highlighted just a few issues, please go through them.

Also, pay attention to linter warnings, they may seem irrelevant now but in larger projects they may cause you a lot of trouble if not solved

EdPirro (approved):
Great job, @HigorPedroso! Keep it up! 🚀

PR #650, opened 2024-05-13
task solution

mykhailonl:
[DEMO LINK](https://mykhailonl.github.io/react_autocomplete/)


src/App.tsx
```diff
@@ -1,73 +1,109 @@
...
+    name: '',
+    born: 0,
+    died: 0,
+    slug: '',
+    fatherName: '',
+    motherName: '',
+    sex: 'm',
+  };
+
+  const [selectedPerson, setSelectedPerson] = useState<Person>(initialPerson);
```
TarasHoliuk, line 20:
```suggestion
  const [selectedPerson, setSelectedPerson] = useState<Person | null>(null);
```

and then this below:

```
const { name, born, died } = selectedPerson || {};
```
Because actually, there is no person selected initially

src/App.tsx
```diff
@@ -1,73 +1,109 @@
...
+  const { name, born, died } = selectedPerson;
+
+  const titleText =
+    selectedPerson.name !== initialPerson.name
+      ? `${name} (${born} - ${died})`
+      : 'No selected person';
+
+  const inputRef = useRef<HTMLInputElement>(null);
+
+  const handleDropdownClick = (a: boolean) => {
```
TarasHoliuk, line 33: what is `a`? - write meaningful parameter name

src/components/autocomplete/Autocomplete.tsx
```diff
@@ -0,0 +1,32 @@
+import React from 'react';
```
TarasHoliuk, line 1: no need to create a separate folder for a component with 1 file only (no separate style files or children components, related only to the component, etc.)

But in case you have a folder - it should be named in PascalCase as well (Autocomplete)

src/App.tsx
```diff
@@ -1,73 +1,109 @@
...
+    setTimeout(() => {
+      setIsDropdownActive(a);
+    }, 200);
```
TarasHoliuk, line 36: I guess the timeout is redundant in this case, so you can use `setIsDropdownActive` instead of `handleDropdownClick` below

TarasHoliuk (changes requested):
GJ 👍 
Added a few small comments

src/App.tsx
mykhailonl: Need this timeout to rewrite selected person

etojeDenys (changes requested):
better to move this message inside the dropdown
image

anastasiiavorobiova (approved):
Great work!

PR #660, opened 2024-05-22
add task solution

DenLysiak:
[DEMO LINK](https://DenLysiak.github.io/react_autocomplete/)


volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code
You need to show selected people in the input 
image

This part of the task is incomplete 
image

You need to hide this block when we don't have people

l4st1m0za (approved):
Good job!

PR #664, opened 2024-05-26
add task solution

wasik07:
[DEMO LINK](https://wasik07.github.io/react_autocomplete/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,73 +1,105 @@
-import React from 'react';
 import './App.scss';
+import { useState, useCallback, useEffect, useRef, useMemo } from 'react';
+import cn from 'classnames';
+import debounce from 'lodash.debounce';
 import { peopleFromServer } from './data/people';
+import { Person } from './types/Person';
+import { PeopleList } from './components/PeopleList';
+
+const DELAY = 300;
 
 export const App: React.FC = () => {
-  const { name, born, died } = peopleFromServer[0];
+  const dropRef = useRef<HTMLDivElement>(null);
+
+  const [selected, setSelected] = useState<Person | null>(null);
+  const [query, setQuery] = useState('');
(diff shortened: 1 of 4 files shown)


src/App.tsx
```diff
@@ -1,73 +1,105 @@
...
+const DELAY = 300;
 
 export const App: React.FC = () => {
-  const { name, born, died } = peopleFromServer[0];
+  const dropRef = useRef<HTMLDivElement>(null);
+
+  const [selected, setSelected] = useState<Person | null>(null);
+  const [query, setQuery] = useState('');
+  const [appliedQuery, setAppliedQuery] = useState('');
+  const [touched, setTouched] = useState(false);
```
TarasHoliuk, line 17:
```suggestion
  const [isTouched, setIsTouched] = useState(false);
```
for boolean variables
but `isTouched` is not the best name here, because we set if false once input is not focused. `isFocused` name explains it better

src/App.tsx
```diff
@@ -1,73 +1,105 @@
...
+
+  // eslint-disable-next-line react-hooks/exhaustive-deps
+  const applyQuery = useCallback(debounce(setAppliedQuery, DELAY), []);
+
+  const normalizeQuery = appliedQuery.trim().toLowerCase();
```
TarasHoliuk, line 22: More just FYI:
It's not a big deal here, but `normalizedQuery` will be created again on every render since it's not memorized, so the same for the `filteredPeople`.
In such case, it's better to normalize it inside `filteredPeople` useMemo's callback, and set `appliedQuery` as a dependency.

TarasHoliuk (changes requested):
Nice solution!

Let's also implement this requirement:
image

SanyaBratashchuk (approved):

PR #678, opened 2024-06-11
add task solution

MakhmudKerimov:
[DEMO LINK](https://MakhmudKerimov.github.io/react_autocomplete/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,15 +1,96 @@
-import React from 'react';
+import React, { useState, useEffect, useRef } from 'react';
 import './App.scss';
 import { peopleFromServer } from './data/people';
+import { Person } from './types/Person';
 
-export const App: React.FC = () => {
-  const { name, born, died } = peopleFromServer[0];
+interface AppProps {
+  debounceDelay?: number;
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -1,15 +1,96 @@
...
+  const [filteredSuggestions, setFilteredSuggestions] = useState<
+    typeof peopleFromServer
+  >([]);
+  const [inputFocused, setInputFocused] = useState<boolean>(false);
+  const [showNoSuggestionsMessage, setShowNoSuggestionsMessage] =
+    useState<boolean>(false);
+  const previousSearchTerm = useRef<string>('');
+  const inputRef = useRef<HTMLInputElement>(null);
+
+  useEffect(() => {
```
TarasHoliuk, line 26: it works but it'd look a bit simpler with `debounce` function usage (and 2 states for typed value - one is immediate and the second is delayed and used to filter people)

At least, read about debounced and how to implement it

FYI

TarasHoliuk (changes requested):

vadiimvooo (approved):
Well done

PR #680, opened 2024-06-13
Develop

arthurfujii:
[DEMO LINK](https://arthurfujii.github.io/react_autocomplete/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -7,7 +7,7 @@ export const App: React.FC = () => {
   const [selectedPerson, setSelectedPerson] = useState<Person | null>(null);
   const showPersonInfo =
     selectedPerson !== null
-      ? `${selectedPerson.name} (${selectedPerson.born}-${selectedPerson.died})`
+      ? `${selectedPerson.name} (${selectedPerson.born} - ${selectedPerson.died})`
       : 'No selected person';
 
   return (
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,73 +1,22 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
-import { peopleFromServer } from './data/people';
+import { Autocomplete } from './components/Autocomplete';
+import { Person } from './types/Person';
 
 export const App: React.FC = () => {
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -1,73 +1,22 @@
...
+import { Autocomplete } from './components/Autocomplete';
+import { Person } from './types/Person';
 
 export const App: React.FC = () => {
-  const { name, born, died } = peopleFromServer[0];
+  const [selectedPerson, setSelectedPerson] = useState<Person | null>(null);
+  const showPersonInfo =
+    selectedPerson !== null
+      ? `${selectedPerson.name} (${selectedPerson.born} - ${selectedPerson.died})`
+      : 'No selected person';
```
alysonvilela, line 11: Its not wrong, but you might not need to verify if its null, just use JS negative

```js
const personInfo = !selectedPerson ? 'No selected person' : `${selectedPerson.name} (${selectedPerson.born} - ${selectedPerson.died})`
```

src/App.tsx
```diff
@@ -1,73 +1,22 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
-import { peopleFromServer } from './data/people';
+import { Autocomplete } from './components/Autocomplete';
+import { Person } from './types/Person';
 
 export const App: React.FC = () => {
-  const { name, born, died } = peopleFromServer[0];
+  const [selectedPerson, setSelectedPerson] = useState<Person | null>(null);
```
alysonvilela, line 7: Your code is becoming professional with right usage of hooks and typescript. Here is a tip: always on selecting some data you can usa identifiers to make as selectable and make traffic between ids and not filled objects.

https://react.dev/learn/choosing-the-state-structure#avoid-duplication-in-state

Note: always think that when
1. You fetched a data from server
2. You want to use one of its data

So you can make reference for id and what is your selectedPerson will become more like

```js
const [selectedPersonId, setSelectedPersonId] = useState<string | null>(null)
const selectedPerson = data.find((person) => person.id === selectedPersonId)
```

Please read the docs I sent and you'll understand it better

alysonvilela (approved):

PR #687, opened 2024-06-26
Task solution

Poviakalo:
[DEMO LINK](https://Poviakalo.github.io/react_autocomplete/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,73 +1,110 @@
-import React from 'react';
+import React, {
+  useState,
+  useEffect,
+  useRef,
+  useMemo,
+  useCallback,
+} from 'react';
+import debounce from 'lodash.debounce';
+
+import 'bulma';
 import './App.scss';
 import { peopleFromServer } from './data/people';
+import { DropdownMenu } from './DropdownMenu';
+import { Person } from './types/Person';
 
 export const App: React.FC = () => {
-  const { name, born, died } = peopleFromServer[0];
+  const [person, setPerson] = useState<Person | null>(null);
+
+  const [text, setText] = useState('');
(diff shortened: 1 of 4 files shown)


src/App.tsx
```diff
@@ -1,73 +1,110 @@
...
+  }, [appliedQuery]);
 
   return (
     <div className="container">
       <main className="section is-flex is-flex-direction-column">
         <h1 className="title" data-cy="title">
-          {`${name} (${born} - ${died})`}
+          {person
+            ? `${person.name} (${person.born} - ${person.died})`
+            : 'No selected person'}
```
volodymyr-soltys97, line 72: Move this logic to the helper variable and use it here

src/DropdownMenu.tsx
```diff
@@ -0,0 +1,33 @@
...
+      <div className="dropdown-content">
+        {list.map(user => {
+          return (
+            <div
+              key={user.name}
+              className="dropdown-item"
+              data-cy="suggestion-item"
+            >
+              <p
+                className={`has-text-${user.sex === 'm' ? 'link' : 'danger'}`}
```
volodymyr-soltys97, line 22: Use the classnames library for add class with condition

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's make your code better
You need to fix this failed test 
image

Need to change title when we edit person in field
image

anastasiia-tilikina (changes requested):
We still need to fix this point
image

Poviakalo:
Please, check again that link.

lerastarynets (approved):
Keep up the great work!

PR #693, opened 2024-07-01
Srtrace/react autocomplete solution

SrTrace:
[DEMO LINK](https://srtrace.github.io/react_autocomplete/)

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,10 +15,11 @@ that will suggest people matching an entered text.
 - when the selected person is displayed in the title, but the value in the input changes, the selected person should be cleared and `No selected person` should be shown.
 
 ## Instructions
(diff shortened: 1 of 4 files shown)


lerastarynets (changes requested):
Selected person don't appear in autocomplete field, pls fix
Screenshot at Jul 01 15-18-10

anastasiiavorobiova (changes requested):
Great work! Consider hiding the list when a user clicks outside the autocomplete. Also, there is no need to show the empty list when there are no suggestions 
Screenshot 2024-07-01 at 16 42 38

etojeDenys (approved):
lgtm

PR #694, opened 2024-07-04
solution

getchards:
[DEMO LINK](https://getchards.github.io/react_autocomplete/)


volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code
1. You need to show selected people in the input 
image

2. Also, you need to fix the failed tests, feel free in the fe_chat

3. Hide this block when we don't have people
image

volodymyr-soltys97 (changes requested):
Hi, this comment still not fixed from the previous review 
image
image

Also, you need to add this data attribute for tests 
image

lerastarynets (changes requested):
Pls try to pass all tests, if your tests pass locally, but you encounter a problem in server tests, pls attach a screen of all local tests passing
Pls ask for help in **fe_chat** if you don't understand something

getchards:
так і не вдалось виправити тести  - звертався по допомогу в чаті

src/App.tsx
```diff
@@ -1,72 +1,107 @@
...
+
+  const handleChoose = (person: Person) => {
+    setCurrentPerson(person);
+    setSearchPart(person.name);
+    setIsDisplayed(false);
+
+    setTimeout(() => {
+      setSuggestedPeople(filterPeople(''));
+      setIsDisplayed(false);
+    }, 0);
```
TarasHoliuk, line 44:
```suggestion
    setIsDisplayed(false);
```
You definitely shouldn't do it `setIsDisplayed(false)` twice
I guess you can just delete `setTimeout`. In case it doesn't work anymore just leave timeout and do `setIsDisplayed` once inside timeout

setSuggestedPeople(filterPeople('')); seems redundant to me here - next time you will show list of people you will redefine it anyway

TarasHoliuk (changes requested):
Let's add a small change to the code as discussed in the chat 🙂

TarasHoliuk (changes requested):
Technical request - to sync with MA platform

TarasHoliuk (approved):

PR #697, opened 2024-07-06
Solution

kyrylomanko:
[DEMO LINK](https://kyrylomanko.github.io/react_autocomplete/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,73 +1,111 @@
-import React from 'react';
+import React, { useMemo, useState } from 'react';
 import './App.scss';
 import { peopleFromServer } from './data/people';
+import debounce from 'lodash.debounce';
+import { Person } from './types/Person';
 
 export const App: React.FC = () => {
-  const { name, born, died } = peopleFromServer[0];
+  const [inputValue, setInputValue] = useState('');
+  const [suggestions, setSuggestions] = useState<Person[]>([]);
+  const [selectedPerson, setSelectedPerson] = useState<Person | null>(null);
+  const [isDropdownVisible, setIsDropdownVisible] = useState(false);
+
+  const debouncedSuggestions = useMemo(
+    () =>
+      debounce((value: string) => {
+        const filteredSuggestions = value
+          ? peopleFromServer.filter(person => person.name.includes(value))
+          : peopleFromServer;
+
+        setSuggestions(filteredSuggestions);
+      }, 300),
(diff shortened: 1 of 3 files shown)


src/App.tsx
```diff
@@ -1,73 +1,111 @@
...
+            {`${selectedPerson.name} (${selectedPerson.born} - ${selectedPerson.died})`}
+          </h1>
+        ) : (
+          <h1 className="title" data-cy="title">
+            No selected person
+          </h1>
+        )}
 
-        <div className="dropdown is-active">
+        <div className={`dropdown ${isDropdownVisible ? 'is-active' : ''}`}>
```
volodymyr-soltys97, line 59: Use the classnames library for add classes with condition

src/App.tsx
```diff
@@ -1,73 +1,111 @@
...
-        </h1>
+        {selectedPerson ? (
+          <h1 className="title" data-cy="title">
+            {`${selectedPerson.name} (${selectedPerson.born} - ${selectedPerson.died})`}
+          </h1>
+        ) : (
+          <h1 className="title" data-cy="title">
+            No selected person
+          </h1>
+        )}
```
volodymyr-soltys97, line 57: You don't need to duplicate this tag twice, you need to put the check inside the tag. It would also be better to save this check in a separate variable and use it here
```suggestion
        {selectedPerson ? (
          <h1 className="title" data-cy="title">
            {`${selectedPerson.name} (${selectedPerson.born} - ${selectedPerson.died})`}
          </h1>
        ) : (
          <h1 className="title" data-cy="title">
            No selected person
          </h1>
        )}
```

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code
The search should not be case sensitive
image
image

src/App.tsx
kyrylomanko: Hi! I don't know why it is casesensitive and not sure how to fix that, I would appreciate some hint cause I don't really see where the problem comes from. I've fixed all other stuff.

anastasiiavorobiova (changes requested):
Great work! Consider fixing the previous comments 
Screenshot 2024-07-07 at 13 15 56
Also, consider hiding the list if the user clicks outside the autocomplete

volodymyr-soltys97 (approved):
Great work 🔥

PR #707, opened 2024-07-18
My solution

Karanelus:
[DEMO LINK](https://karanelus.github.io/react_autocomplete/)

### DIFF
--- a/src/components/dropdownList/DropdownList.tsx
+++ b/src/components/dropdownList/DropdownList.tsx
@@ -1,0 +1,72 @@
+import { useEffect, useRef } from 'react';
+import { Person } from '../../types/Person';
+import DropdownItem from '../dropdownItem/DropdownItem';
+
+type Props = {
+  people: Person[];
+  onClick: (person: Person) => void;
+  isFocused: boolean;
+  onFocusElement: (state: boolean) => void;
+  inputValue: string;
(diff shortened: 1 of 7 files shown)


src/components/dropdownList/DropdownList.tsx
```diff
@@ -0,0 +1,72 @@
...
+          onChange={onChangeInput}
+          onFocus={() => onFocusElement(true)}
+        />
+
+        {!!people.length && isFocused && (
+          <div className="dropdown-menu" role="menu" data-cy="suggestions-list">
+            <div className="dropdown-content">
+              {people.map(person => (
+                <DropdownItem
+                  key={person.name}
```
natalia-klonowska, line 59: person name is not unique enough use slug instead ;)

natalia-klonowska (changes requested):

src/components/dropdownList/DropdownList.tsx
```diff
@@ -0,0 +1,72 @@
...
+              {people.map(person => (
+                <DropdownItem
+                  key={person.slug}
+                  person={person}
+                  onClick={onClick}
+                />
+              ))}
+            </div>
+          </div>
+        )}
```
Zibi95, line 66: Think of a way how would you get rid of that prop drilling here.

Zibi95 (approved):
Well done 🥇

PR #722, opened 2024-08-12
add solution

RimmaBuzuluk:
[DEMO LINK](https://RimmaBuzuluk.github.io/react_autocomplete/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,17 +1,60 @@
-import React from 'react';
+import React, { useEffect, useMemo, useState } from 'react';
 import './App.scss';
 import { peopleFromServer } from './data/people';
+import { Person } from './types/Person';
 
-export const App: React.FC = () => {
-  const { name, born, died } = peopleFromServer[0];
+interface AppProps {
+  debounceDelay?: number;
+}
+
+export const App: React.FC<AppProps> = ({ debounceDelay = 300 }) => {
+  const people: Person[] = [...peopleFromServer];
+  const [isSelected, setIsSelected] = useState(false);
+  const [selectedPerson, setSelectedPerson] = useState<Person>(people[0]);
+  const [filterUserText, setFilterUserText] = useState('');
+  const [debouncedText, setDebouncedText] = useState(filterUserText);
+  const { name, born, died } = selectedPerson;
+
+  const setUserText = (event: string) => {
+    setFilterUserText(event);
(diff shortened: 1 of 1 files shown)


YaroslavKolesnyk (changes requested):
Hi, sorry but your link is not working😓, need to fix it, overall well written code👍

YaroslavKolesnyk (approved):

src/App.tsx
```diff
@@ -1,73 +1,103 @@
...
-export const App: React.FC = () => {
-  const { name, born, died } = peopleFromServer[0];
+interface AppProps {
+  debounceDelay?: number;
+}
+
+export const App: React.FC<AppProps> = ({ debounceDelay = 300 }) => {
+  const people: Person[] = useMemo(
+    () => [...peopleFromServer],
+    [peopleFromServer],
```
ArturErmolenko, line 13: you shouldn't use useMemo here cause 'peopleFromServer' aren't valid dependencies because mutating them doesn't re-render the component

src/App.tsx
```diff
@@ -1,73 +1,103 @@
...
+  };
+
+  useEffect(() => {
+    const handler = setTimeout(() => {
+      setDebouncedText(filterUserText);
+    }, debounceDelay);
+
+    return () => {
+      clearTimeout(handler);
+    };
```
ArturErmolenko, line 33: 🔥

src/App.tsx
```diff
@@ -1,73 +1,103 @@
...
-
+        {isSelected ? (
+          <h1 className="title" data-cy="title">
+            {`${name} (${born} - ${died})`}
+          </h1>
+        ) : (
+          <h1 className="title" data-cy="title">
+            No selected person
+          </h1>
+        )}
```
ArturErmolenko, line 58: h1 is the same in each case, so can we change only content?

src/App.tsx
```diff
@@ -1,73 +1,103 @@
...
+                  <p className="has-text-link">{person.name}</p>
+                </div>
+              ))}
             </div>
           </div>
         </div>
 
-        <div
-          className="
+        {filterPeople.length === 0 && (
```
ArturErmolenko, line 86:
```suggestion
        {!filterPeople.length  && (
```

ArturErmolenko (approved):
Well done 👍 
Pay attention on comments

PR #749, opened 2024-09-24
solution

Shevchuchka:
https://Shevchuchka.github.io/react_autocomplete/

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,8 +1,6 @@
 # React Autocomplete
 
-You have an array of people. Implement the Autocomplete component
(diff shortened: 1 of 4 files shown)


volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code
You need to hide people when input isn't focused 
image

Hide this block when we don't have people 
image

volodymyr-soltys97 (changes requested):
Hi, this comment still not fixed from the previous review 
image

volodymyr-soltys97 (approved):
Good job 👍

PR #754, opened 2024-10-07
solved task

PikalovAndrey:
[DEMO LINK](https://PikalovAndrey.github.io/react_autocomplete/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -25,13 +25,11 @@ export const App: React.FC = () => {
   };
 
   const handleChangeText = (event: React.ChangeEvent<HTMLInputElement>) => {
-    const value = event.target.value;
+    setFilterValue(event.target.value);
 
-    setFilterValue(value);
+    debouncedSetFilterValue(event.target.value);
 
-    debouncedSetFilterValue(value);
-
-    if (selectedPerson && value) {
+    if (selectedPerson && event.target.value) {
       setSelectedPerson(null);
     }
 
@@ -51,13 +49,11 @@ export const App: React.FC = () => {
   return (
     <div className="container">
       <main className="section is-flex is-flex-direction-column">
-        {selectedPerson && (
+        {selectedPerson ? (
           <h1 className="title" data-cy="title">
             {`${selectedPerson?.name} (${selectedPerson?.born} - ${selectedPerson?.died})`}
           </h1>
-        )}
-
-        {!selectedPerson && (
+        ) : (
           <h1 className="title" data-cy="title">
             No selected person
           </h1>
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,15 +1,21 @@
-import React, { useMemo, useState } from 'react';
+import React, { useMemo, useState, useCallback } from 'react';
 import './App.scss';
 import { peopleFromServer } from './data/people';
 import { Person } from './types/Person';
 import classNames from 'classnames';
+import debounce from 'lodash.debounce';
 
 export const App: React.FC = () => {
   const [selectedPerson, setSelectedPerson] = useState<Person | null>(null);
-  const [typingTimeout, setTypingTimeout] = useState<number | undefined>(
-    undefined,
-  );
   const [filterValue, setFilterValue] = useState('');
+  const [debouncedFilterValue, setDebouncedFilterValue] = useState(filterValue);
+
+  const debouncedSetFilterValue = useCallback(
+    debounce((value: string) => {
+      setDebouncedFilterValue(value);
+    }, 300),
+    [],
+  );
 
   const handleSelectedPerson = (person: Person | null) => {
     setSelectedPerson(person);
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -1,15 +1,52 @@
...
+
+  const [filterValue, setFilterValue] = useState('');
+
+  const handleSelectedPerson = (person: Person | null) => {
+    setSelectedPerson(() => person);
+    setFilterValue('');
+  };
+
+  const handleChangeText = (event: React.ChangeEvent<HTMLInputElement>) => {
+    const value = event.target.value;
```
KrivorotkoYurii: You can also put event.target.value straight to your function 
```suggestion
    setFilterValue(event.target.value);
```

src/App.tsx
```diff
@@ -18,56 +55,32 @@ export const App: React.FC = () => {
...
-              </div>
+              {filteredPeople.length > 0 ? (
+                filteredPeople.map(person => (
+                  <div
+                    className="dropdown-item"
+                    data-cy="suggestion-item"
+                    onClick={() => handleSelectedPerson(person)}
+                    key={person.slug}
+                  >
+                    <p className="has-text-danger">{person.name}</p>
```
KrivorotkoYurii: as it was on the start version you need you put different text colors for men and weman

src/App.tsx
```diff
@@ -1,15 +1,52 @@
-import React from 'react';
+import React, { useMemo, useState } from 'react';
 import './App.scss';
 import { peopleFromServer } from './data/people';
+import { Person } from './types/Person';
 
 export const App: React.FC = () => {
-  const { name, born, died } = peopleFromServer[0];
+  const [selectedPerson, setSelectedPerson] = useState<Person | null>(null);
+  const [typingTimeout, setTypingTimeout] = useState<number | undefined>(
```
KrivorotkoYurii: According to the check-list suggestions should appear after some delay in typing (debounce); So you need to use debounce function ( or your own function  or import debounce from  'lodash.debounce'

KrivorotkoYurii: According to the check-list the delay should be customizable via props (default value is 300ms);

src/App.tsx
```diff
@@ -18,56 +55,32 @@ export const App: React.FC = () => {
               type="text"
```
KrivorotkoYurii, line 67: According to the check-list  - save selected suggestion text to the input on click and close the list;

KrivorotkoYurii (changes requested):
Check the check- list

KrivorotkoYurii (approved):

src/App.tsx
VolodymyrKirichenko: good point 🔥

src/App.tsx
```diff
@@ -18,56 +55,41 @@ export const App: React.FC = () => {
...
-
-              <div className="dropdown-item" data-cy="suggestion-item">
-                <p className="has-text-danger">Petronella de Decker</p>
-              </div>
-
-              <div className="dropdown-item" data-cy="suggestion-item">
-                <p className="has-text-danger">Elisabeth Hercke</p>
-              </div>
+              {filteredPeople.length !== 0 &&
+                !selectedPerson &&
```
VolodymyrKirichenko: will be the same. use this syntax in all such cases
```suggestion
              {(filteredPeople.length && !selectedPerson) && (
```

src/App.tsx
```diff
@@ -18,56 +55,41 @@ export const App: React.FC = () => {
...
+                      className={classNames({
+                        'has-text-danger': person.sex === 'f',
+                        'has-text-link': person.sex === 'm',
+                      })}
+                    >
+                      {person.name}
+                    </p>
+                  </div>
+                ))}
+              {filteredPeople.length === 0 && (
```
VolodymyrKirichenko: the same
```suggestion
              {!filteredPeople.length && (
```

VolodymyrKirichenko (changes requested):
ur selection should work differently.
1) It must be closed at the beginning
2) We should be able to reopen it without typing. Use `onFocus` attribute to track this
[Check ur bug](https://www.loom.com/share/52b0964d86c54a0bac4dd3080334ea3e)

PikalovAndrey (commented):
if I understood correctly. this is the best option

[DEMO LINK](https://PikalovAndrey.github.io/react_autocomplete/)

src/App.tsx
VolodymyrKirichenko: u don't need a separate variable for it. u used it only once.

src/App.tsx
```diff
@@ -1,73 +1,123 @@
...
+          <h1 className="title" data-cy="title">
+            {`${selectedPerson?.name} (${selectedPerson?.born} - ${selectedPerson?.died})`}
+          </h1>
+        )}
+
+        {!selectedPerson && (
+          <h1 className="title" data-cy="title">
+            No selected person
+          </h1>
+        )}
```
VolodymyrKirichenko: It will be shortest
```suggestion
        {selectedPerson ? (
          <h1 className="title" data-cy="title">
            {`${selectedPerson?.name} (${selectedPerson?.born} - ${selectedPerson?.died})`}
          </h1>
        ) : (
          <h1 className="title" data-cy="title">
            No selected person
          </h1>
        )}
```

VolodymyrKirichenko (approved):
Good job 👍

PR #766, opened 2024-10-27
add task solution

BohdanDymydiuk:
[DEMO LINK](https://BohdanDymydiuk.github.io/react_autocomplete/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,4 +1,4 @@
-import React, { useMemo, useRef, useState } from 'react';
+import React, { useEffect, useMemo, useRef, useState } from 'react';
 import { Autocomplete } from './components/Autocomplete';
 import './App.scss';
 import { peopleFromServer } from './data/people';
@@ -20,13 +20,13 @@ export const App: React.FC = () => {
     }, delay);
   }
 
-  window.addEventListener('click', () => {
-    setFocus(false);
-  });
-
   const filteredPeople = useMemo(() => {
     return peopleFromServer.filter(person => person.name.includes(query));
   }, [query]);
+
+  useEffect(() => {
+    setFocus(false);
+  }, [selectedPerson]);
 
   return (
     <div className="container">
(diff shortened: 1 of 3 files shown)


src/App.tsx
```diff
@@ -1,73 +1,65 @@
...
 import { peopleFromServer } from './data/people';
+import { Person } from './types/Person';
 
 export const App: React.FC = () => {
-  const { name, born, died } = peopleFromServer[0];
+  const [selectedPerson, setSelectedPerson] = useState<Person | null>(null);
+  const { name, born, died } = selectedPerson || {};
+  const [delay] = useState(300);
+  const [query, setQuery] = useState('');
+  const [focus, setFocus] = useState(false);
```
VitaliyBondarenko1982: App component shouldn't know about query or focus. move logic with it to Autocomplete component.

src/App.tsx
```diff
@@ -1,73 +1,65 @@
...
+import { Autocomplete } from './components/Autocomplete';
 import './App.scss';
 import { peopleFromServer } from './data/people';
+import { Person } from './types/Person';
 
 export const App: React.FC = () => {
-  const { name, born, died } = peopleFromServer[0];
+  const [selectedPerson, setSelectedPerson] = useState<Person | null>(null);
+  const { name, born, died } = selectedPerson || {};
+  const [delay] = useState(300);
```
VitaliyBondarenko1982: why do you need to keep delay in state? it can be just constant outside of component.

VitaliyBondarenko1982 (changes requested):
Good job.
Needs some improvements

Check comments and screencast.
- filtered suggestion should be case insensitive
- after selection not works filtering and dropdown not wokks properly

https://github.com/user-attachments/assets/e1c5cad5-5925-4ff4-8fe2-4007cbe1b05d

maxim2310 (approved):
Well done!!!

PR #770, opened 2024-11-01
Solution

ab3MN:
[DEMO LINK](https://ab3MN.github.io/react_autocomplete/)

### DIFF
--- a/src/components/Autocomplete.tsx
+++ b/src/components/Autocomplete.tsx
@@ -1,0 +1,102 @@
+import {
+  ChangeEvent,
+  FC,
+  ReactNode,
+  useCallback,
+  useMemo,
+  useState,
+} from 'react';
+import { Person } from '../types/Person';
+
+import _debounce from 'lodash/debounce';
+
+interface IProps {
+  people: Person[];
+  onSelected: (person: Person | null) => void;
+  selectedPerson: Person | null;
+  delay?: number;
+}
+
+const Autocomplete: FC<IProps> = ({
+  people,
+  onSelected,
+  selectedPerson,
+  delay = 300,
+}): ReactNode => {
+  const [isActive, setActive] = useState(false);
+  const [query, setQuery] = useState('');
+  const [debounceQuery, setDebounceQuery] = useState('');
+
+  const debouncedSetQuery = _debounce(setDebounceQuery, delay);
+
+  const handleChange = (event: ChangeEvent<HTMLInputElement>) => {
+    const { value } = event.target;
+
+    setActive(true);
+    setQuery(value);
+    debouncedSetQuery(value);
+    onSelected(null);
+  };
+
+  const handleSelectPerson = (person: Person) => {
+    onSelected(person);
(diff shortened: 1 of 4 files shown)


src/components/Autocomplete.tsx
```diff
@@ -0,0 +1,95 @@
+import { ChangeEvent, FC, ReactNode, useMemo, useState } from 'react';
+import { Person } from '../types/Person';
+
+import _debounce from 'lodash.debounce';
+
+interface IProps {
```
VenomMB: AutocompleteProps

src/components/Autocomplete.tsx
```diff
@@ -0,0 +1,95 @@
...
+            is-align-self-flex-start
+          "
+                role="alert"
+                data-cy="no-suggestions-message"
+              >
+                <p className="has-text-danger">No matching suggestions</p>
+              </div>
+            )}
+          </div>
+        </div>
```
VenomMB: move to separate component

src/components/Autocomplete.tsx
```diff
@@ -0,0 +1,95 @@
...
+
+            {filtredPeople.length === 0 && (
+              <div
+                className="
+            notification
+            is-danger
+            is-light
+            mt-3
+            is-align-self-flex-start
+          "
```
VenomMB:
```suggestion
                  notification
                  is-danger
                  is-light
                  mt-3
                  is-align-self-flex-start
              "
```

VenomMB (changes requested):
Great job, let's work on improvements now

src/components/Autocomplete.tsx
ab3MN: Could you please explain,why IProps  is bad name?I read an article where was written that  it's a good practice to give shortly name like IProps or TProps to components props types.There was also written that good practice is to start a name with "I" or "T" depending on the type,but I see that in Mate Acedamy don't start name with"I" or "T".

src/components/Autocomplete.tsx
ab3MN: Could you advise some prettier config or something like prettier for it?I wasn't paying attention on it while codding,it would be good if it do automatically

Moroz-Dmytro (approved):
LGTM 👍

src/components/Autocomplete.tsx
Moroz-Dmytro: Screenshot 2024-11-05 at 11 43 37

@ab3MN It doesn't work?

src/components/Autocomplete.tsx
Moroz-Dmytro: AutocompleteProps just better explains what this interface is for

src/components/Autocomplete.tsx
ab3MN: @Moroz-Dmytro no:( it's worked only for variables,but not for classNames,perhaps classNames is a string and tabsize isn't worked in string

PR #793, opened 2024-12-07
Changes

Mykhalishere:
[DEMO LINK](https://Mykhalishere.github.io/react_autocomplete/)


src/App.tsx
```diff
@@ -1,15 +1,80 @@
...
+  useMemo,
+  useRef,
+  useState,
+} from 'react';
 import './App.scss';
 import { peopleFromServer } from './data/people';
+import { Person } from './types/Person';
+import debounce from 'lodash.debounce';
+
+function filterPerson(arr: Person[], query: string) {
```
TarasHoliuk, line 13:
```suggestion
function filterPeople(people: Person[], query: string) {
```
arr is a bad name for variable/parameter - you should never use it

src/App.tsx
```diff
@@ -1,15 +1,80 @@
...
+  return query ? [...arr].filter(elem => elem.name.includes(query)) : [...arr];
+}
 
 export const App: React.FC = () => {
-  const { name, born, died } = peopleFromServer[0];
+  const [query, setQuery] = useState('');
+  const [appliedQuery, setAppliedQuery] = useState('');
+  const [title, setTitle] = useState('No selected person');
+
+  const [showArray, setShowArray] = useState(false);
```
TarasHoliuk, line 22: What is Array? Do not use such names

src/App.tsx
```diff
@@ -1,15 +1,80 @@
...
+  }, [appliedQuery]);
+
+  // eslint-disable-next-line react-hooks/exhaustive-deps
+  const applyQuery = useCallback(debounce(setAppliedQuery, 300), []);
+
+  const handleChangeInputName = (
+    event: React.ChangeEvent<HTMLInputElement>,
+  ) => {
+    setQuery(event.target.value);
+    setTitle('No selected person');
```
TarasHoliuk, line 36: you may create a variable DEFAULT_TITLE with this string outside of the component and reuse it

src/App.tsx
```diff
@@ -1,15 +1,80 @@
...
 export const App: React.FC = () => {
-  const { name, born, died } = peopleFromServer[0];
+  const [query, setQuery] = useState('');
+  const [appliedQuery, setAppliedQuery] = useState('');
+  const [title, setTitle] = useState('No selected person');
+
+  const [showArray, setShowArray] = useState(false);
+  const showMessageRef = useRef(false);
+
+  const updatedPersons = useMemo(() => {
```
TarasHoliuk, line 25: filteredPeople variable name will describe it better (they aren't updated)

src/App.tsx
```diff
@@ -1,15 +1,80 @@
...
+      setShowArray(false);
+      showMessageRef.current = true;
+    }
+  };
+
+  useEffect(() => {
+    if (appliedQuery && updatedPersons.length > 0) {
+      const { name, born, died } = updatedPersons[0];
+
+      setTitle(`${name} (${born} - ${died})`);
```
TarasHoliuk, line 60: You should set it on click, not when user type something in the input

src/App.tsx
```diff
@@ -50,24 +133,27 @@ export const App: React.FC = () => {
...
+              </div> */}
               </div>
             </div>
-          </div>
+          )}
         </div>
 
-        <div
-          className="
+        {showMessageRef.current && (
```
TarasHoliuk, line 142: you use it for conditional rendering - so it should be state, because the ref value change doesn't cause rerender
Even if other state values change at the same time and it will work you shouldn't do it like this

TarasHoliuk (changes requested):
GJ 👍 
Pay attention to naming

PR #795, opened 2024-12-09
solution

maxsabo:
[DEMO LINK](https://maxsabo.github.io/react_autocomplete/)


LiliiaVol (changes requested):
Your code look good, but something went wrong with "no matching suggestions", and also I think your input have to be cleaned value when person is chosen

PR #833, opened 2025-01-24
Solution

pdabizha:
[DEMO LINK](https://pdabizha.github.io/react_autocomplete/)

### DIFF
--- a/src/components/DropDownItem/DropDownItem.tsx
+++ b/src/components/DropDownItem/DropDownItem.tsx
@@ -1,0 +1,23 @@
+import React from 'react';
+import { Person } from '../../types/Person';
+
+type Props = {
+  option: Person;
+  onSelect: (option: Person) => void;
+};
+
+export const DropDownItem: React.FC<Props> = ({ option, onSelect }) => {
+  const handleClick = () => {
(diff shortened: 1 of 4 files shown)


src/components/DropDownItem/DropDownItem.tsx
```diff
@@ -0,0 +1,23 @@
...
+export const DropDownItem: React.FC<Props> = ({ option, onSelect }) => {
+  const handleClick = () => {
+    onSelect(option); // Передаём выбранную опцию вверх
+  };
+
+  return (
+    <div
+      className="dropdown-item"
+      data-cy="suggestion-item"
+      onClick={handleClick}
```
etojeDenys, line 18: I think you want to use here the onMouseDown event

etojeDenys (changes requested):
looks like it doesn't work
https://github.com/user-attachments/assets/60a3ed02-a0a0-4580-8011-b638d8744034

src/components/DropDownItem/DropDownItem.tsx
pdabizha: Yes, I fixed this. Is there are any other issues? Should I re-request review?

PR #838, opened 2025-01-27
Solution

djtyrf312:
[DEMO LINK](https://djtyrf312.github.io/react_autocomplete/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -10,12 +10,13 @@ export const App: React.FC = () => {
     null,
   );
   const [appliedQuery, setAppliedQuery] = React.useState('');
-  const filteredPeople = React.useMemo(() => {
-    return peopleFromServer.filter(person =>
-      person.name.toLowerCase().includes(appliedQuery.toLowerCase()),
-    );
-  }, [appliedQuery]);
-  const isEmptyPeopleList = filteredPeople.length === 0;
+  const people = React.useMemo(
+    () =>
+      peopleFromServer.filter(person =>
+        person.name.toLowerCase().includes(appliedQuery.toLowerCase()),
+      ),
+    [appliedQuery],
+  );
 
   return (
     <div className="container">
@@ -29,11 +30,10 @@ export const App: React.FC = () => {
         <Dropdown
           onQueryChange={setAppliedQuery}
           onPersonSelect={setSelectedPerson}
-          people={filteredPeople}
-          isEmptyPeople={isEmptyPeopleList}
+          people={people}
           debounceDelay={300}
         />
-        {isEmptyPeopleList && <Alert />}
+        {people.length === 0 && <Alert />}
       </main>
     </div>
   );
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -2,34 +2,50 @@ import React from 'react';
 import './App.scss';
 import { peopleFromServer } from './data/people';
 import { Person } from './types/Person';
+import debounce from 'lodash.debounce';
 
 export const App: React.FC = () => {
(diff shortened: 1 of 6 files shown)


src/App.tsx
```diff
@@ -1,73 +1,39 @@
 import React from 'react';
```
naziamoff, line 1: you can import hooks right away so you can use them right away instead of `React.hookName`
```suggestion
import React, { useState, useMemo } from 'react';
```

src/App.tsx
```diff
@@ -1,73 +1,39 @@
...
-  const { name, born, died } = peopleFromServer[0];
+  const [selectedPerson, setSelectedPerson] = React.useState<Person | null>(
+    null,
+  );
+  const [appliedQuery, setAppliedQuery] = React.useState('');
+  const filteredPeople = React.useMemo(() => {
+    return peopleFromServer.filter(person =>
+      person.name.toLowerCase().includes(appliedQuery.toLowerCase()),
+    );
+  }, [appliedQuery]);
```
naziamoff: Since there's no logic inside - you can return right away from the callback
```suggestion
  const filteredPeople = React.useMemo(() => (
    peopleFromServer.filter(person =>
      person.name.toLowerCase().includes(appliedQuery.toLowerCase()),
    );
  ), [appliedQuery]);
```

src/App.tsx
```diff
@@ -1,73 +1,39 @@
...
+import { Alert } from './components/alert/Alert';
+import { Dropdown } from './components/dropdown/Dropdown';
 
 export const App: React.FC = () => {
-  const { name, born, died } = peopleFromServer[0];
+  const [selectedPerson, setSelectedPerson] = React.useState<Person | null>(
+    null,
+  );
+  const [appliedQuery, setAppliedQuery] = React.useState('');
+  const filteredPeople = React.useMemo(() => {
```
naziamoff: Code be like:

image

src/components/dropdownMenu/DropdownMenu.tsx
```diff
@@ -0,0 +1,40 @@
...
+};
+
+export const DropdownMenu: React.FC<Props> = ({
+  people,
+  dropdownRef,
+  setSelectedPerson,
+  setDisplayedDropdown,
+  setQuery,
+}: Props) => {
+  return (
```
naziamoff:
```suggestion
}: Props) => (
  ...the stuff from the return statement goes here
```

src/App.tsx
```diff
@@ -1,73 +1,39 @@
...
+    return peopleFromServer.filter(person =>
+      person.name.toLowerCase().includes(appliedQuery.toLowerCase()),
+    );
+  }, [appliedQuery]);
+  const isEmptyPeopleList = filteredPeople.length === 0;
```
naziamoff: it doesn't seem that you need this variable. The nested components already have access to the variable `people`, so they can check whether it is empty on their own

src/components/dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,73 @@
...
+
+type Props = {
+  onQueryChange: React.Dispatch<React.SetStateAction<string>>;
+  onPersonSelect: React.Dispatch<React.SetStateAction<Person | null>>;
+  people: Person[];
+  isEmptyPeople: boolean;
+  debounceDelay: number;
+};
+
+export const Dropdown: React.FC<Props> = ({
```
naziamoff: You can import `FC` from React as well, so `React.` part can be omitted when using

naziamoff (approved):

PR #841, opened 2025-01-29
add task solution

naz4ik:


src/App.tsx
```diff
@@ -1,72 +1,88 @@
...
+  };
 
   return (
     <div className="container">
       <main className="section is-flex is-flex-direction-column">
         <h1 className="title" data-cy="title">
-          {`${name} (${born} - ${died})`}
+          {selectedPerson
+            ? `${selectedPerson.name} (${selectedPerson.born} - ${selectedPerson.died})`
+            : 'No selected person'}
```
volodymyr-soltys97, line 36: Move this logic from jsx to the helper variable and use it here

src/App.tsx
```diff
@@ -1,72 +1,88 @@
...
         </h1>
 
-        <div className="dropdown is-active">
+        <div className={`dropdown ${isDropdownVisible ? 'is-active' : ''}`}>
```
volodymyr-soltys97, line 39: Use the classnames library for add classes with condition

volodymyr-soltys97 (changes requested):
Hi, there is no demo link, also, you need to add demo link to the PR description 
image

PR #891, opened 2025-03-16
sol

dvdmsk:
- [DEMO LINK](https://dvdmsk.github.io/react_autocomplete/)


src/conponents/DropDown/DropDown.tsx
```diff
@@ -0,0 +1,82 @@
...
+
+    const people = peopleFromServer.filter(person =>
+      person.name.toLowerCase().includes(appliedQuery.trim().toLowerCase()),
+    );
+
+    if (people.length === 0) {
+      setError({ isError: true, errorMessage: 'No matching suggestions' });
+    } else {
+      setError({ isError: false, errorMessage: '' });
+    }
```
etojeDenys, line 37: try to avoid calling setState functions inside the useMemo
you can move this logic in useEffect

etojeDenys (changes requested):
close the dropdown on click and select the person in input value
image

PR #894, opened 2025-03-19
create solution

daniil-trusov:
[DEMO LINK](https://daniil-trusov.github.io/react_autocomplete/


AnOvramenko (approved):
Good job 👍 
But your code has some issues
your dropdown shows even if it is empty
image
Names on the list should be in different colors for men and women.
image

PR #898, opened 2025-03-20
task solution

maponomarenko:
[DEMO LINK](https://maponomarenko.github.io/react_autocomplete/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -37,14 +37,15 @@ export const App: React.FC = () => {
   const makeDropdownInactive = debounce(setIsDropdownActive, 300);
 
   const handleDropdownInactive = () => {
-    if (!dropdownQuery) {
+    if (!dropdownQuery.trim()) {
       makeDropdownInactive(false);
     }
   };
 
+  // eslint-disable-next-line react-hooks/exhaustive-deps
   const applyQuery = useCallback(debounce(setAppliedQuery, 1000), []);
 
-  const handleQueryChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+  const handleInputChange = (event: React.ChangeEvent<HTMLInputElement>) => {
     setDropdownQuery(event.target.value);
     applyQuery(event.target.value);
     setChosenPerson(null);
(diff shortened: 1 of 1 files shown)


AnOvramenko (approved):
Well done!)🥇  
In my opinion, debounce for onBlur is overkill, you can use simple setTimeout)
Also, you can reduce the timer for onBlur event and your dropdown should hide when you select person, now it works only once

src/App.tsx
```diff
@@ -1,73 +1,117 @@
...
+  const visiblePeople = handleFilter(appliedQuery, peopleFromServer);
+
+  const handleDropdownInactive = () => {
+    return setTimeout(() => {
+      setIsDropdownActive(false);
+    }, 300);
+  };
+
+  // eslint-disable-next-line react-hooks/exhaustive-deps
+  const applyQuery = useCallback(debounce(setAppliedQuery, 1000), []);
```
alexander-ignatow, line 44: 1. 1000 is too long for the text inputs. The usual value is 250-300 milliseconds
2. Here is the hint on how to fix the linter warning: https://kyleshevlin.com/debounce-and-throttle-callbacks-with-react-hooks/
3. Per task - this delay should be customizable

src/App.tsx
```diff
@@ -1,73 +1,117 @@
...
+  const [appliedQuery, setAppliedQuery] = useState('');
+
+  const visiblePeople = handleFilter(appliedQuery, peopleFromServer);
+
+  const handleDropdownInactive = () => {
+    return setTimeout(() => {
+      setIsDropdownActive(false);
+    }, 300);
+  };
+
```
alexander-ignatow, line 42: Why do you use any timeout here?
I don't see anything in the task on this - and UI/UX looks little bit weird because of this delay

alexander-ignatow (changes requested):
Good job, but please, read the task more carefully:

> Implement the Autocomplete component

You don't have a separate `Autocomplete` component

This requirement is also missed:
> suggestions should appear after some delay in typing (debounce);
> the delay should be customizable via props (default value is 300ms);

So what is expected, you need to create a separate file (`Autocomplete.tsx`) with Autocomplete component, that has `delay` property (other properties are up to you)

PR #905, opened 2025-03-23
solution

kami2693:
[DEMO LINK](https://kami2693.github.io/react_autocomplete/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,73 +1,25 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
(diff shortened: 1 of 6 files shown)


src/App.tsx
```diff
@@ -1,73 +1,25 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
 import { peopleFromServer } from './data/people';
```
R-Ohirok, line 3: peopleFromServer are not used in App.tsx, it may be better to import them directly into Autocomplete.tsx

R-Ohirok (approved):
Good job!

PR #906, opened 2025-03-25
add task solution

SofiiaPelenska:
[DEMO LINK](https://SofiiaPelenska.github.io/react_autocomplete/)


src/App.tsx
```diff
@@ -1,73 +1,108 @@
...
 import { peopleFromServer } from './data/people';
+import classNames from 'classnames';
+import { Person } from './types/Person';
 
 export const App: React.FC = () => {
-  const { name, born, died } = peopleFromServer[0];
+  const [query, setQuery] = useState('');
+  const [appliedQuery, setAppliedQuery] = useState('');
+  const [selectedPerson, setSelectedPerson] = useState<Person | null>(null);
+  const [isDropDown, setIsDropDown] = useState(true);
```
alexander-ignatow: The dropdown is visible at the start of the app.
This is not normal behavior of the component - I'd suggest showing it on focus and hiding on blur (and possibly on click - you can experiment a bit to get the best UX)

alexander-ignatow (changes requested):
Good job, but please, fix the logic of how the dropdown is shown/hidden

alexander-ignatow (approved):
Good job! 👍

PR #941, opened 2025-05-16
Develop

14lucas-mendes:
[DEMO LINK](https://14lucas-mendes.github.io/react_autocomplete/)


pedro-ruas (changes requested):
You have some errors:
- The message "No matches" is always displaying, even if there is a match
- Clicking on a match should copy the text, which isn't happening

PR #953, opened 2025-05-31
solution

izuwiiii:
[DEMO LINK](https://izuwiiii.github.io/react_autocomplete/)

### DIFF
--- a/src/components/Dropdown/Dropdown.tsx
+++ b/src/components/Dropdown/Dropdown.tsx
@@ -14,6 +14,8 @@ interface DropdownProps {
   setIsChanged: (value: boolean) => void;
   delay: number;
   setAppliedQuery: React.Dispatch<React.SetStateAction<string>>;
+  applyQuery: React.Dispatch<React.SetStateAction<string>>;
+  handleQueryChange: (event: React.ChangeEvent<HTMLInputElement>) => void;
 }
 
 export const Dropdown: React.FC<DropdownProps> = ({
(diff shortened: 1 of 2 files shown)


src/components/Dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,85 @@
...
+  setQuery,
+  field,
+  query,
+  filteredPeople,
+  onSelected,
+  setIsChanged,
+  delay,
+  setAppliedQuery,
+}) => {
+  const applyQuery = useCallback(debounce(setAppliedQuery, delay), []);
```
lidiia-tsymborovych: const applyQuery = useCallback(debounce(setAppliedQuery, delay), [delay]);
debounce should update in case delay will be changed

src/components/Dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,85 @@
...
+
+  const handleQueryChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+    setQuery(event.target.value);
+    applyQuery(event.target.value);
```
lidiia-tsymborovych: applyQuery is always called, even if the value has not changed. You should add some checks :   
const newValue = event.target.value;
  setQuery(newValue);
  if (newValue !== appliedQuery) {
    applyQuery(newValue);
  }

src/components/Dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,85 @@
...
+        <div className="dropdown-menu" role="menu" data-cy="suggestions-list">
+          <div className="dropdown-content">
+            {filteredPeople.map(person => {
+              return (
+                <div
+                  key={person.slug}
+                  className="dropdown-item"
+                  data-cy="suggestion-item"
+                  onClick={() => {
+                    onSelected(person);
```
lidiia-tsymborovych, line 75: save selected suggestion text to the input on click and close the list

lidiia-tsymborovych (changes requested):
Good job, but there are several key issues that need to be fixed😇

PR #962, opened 2025-06-05
add task solution

hannachachacha:
[DEMO LINK](https://hannachachacha.github.io/react_autocomplete/)


hannachachacha:
@mateacademy-ai-mentor Notification is shown when filteredPeople.length === 0. If input is empty, filteredPeople will be equal people (hence peopleFromServer). And peopleFromServer.length isn't 0

PR #1017, opened 2025-08-20
implement autocomplete

YuliaProkipchuk:
[DEMO LINK](https://YuliaProkipchuk.github.io/react_autocomplete/)


isobol-js (approved):
Good job, mate!

YuliaProkipchuk:
@mateacademy-ai-mentor searchText.trim() means that filter works only when i don't have entered spaces only, so this works correctly. Since i have onMouseDown event it means that while clicking on item it does update selected person and closes input. this logic also works correctly, since all test pass. Other changes i've made according to your recommendations and requirements

PR #1047, opened 2025-10-01
add solution

Aiiyuu:
I found it challenging to understand useCallback, useMemo, and useEffect, so I'm not sure if my task is implemented correctly.
Please criticize it as much as possible, and explain what I’m doing wrong if needed. Maybe that will help me understand it better.

[DEMO LINK](https://aiiyuu.github.io/react_autocomplete/)


src/App.tsx
```diff
@@ -1,16 +1,88 @@
...
+  return function (...args: any) {
+    window.clearTimeout(timerId);
+
+    timerId = window.setTimeout(() => {
+      callback(...args);
+    }, delay);
+  };
+}
+
+const DEBOUNCE_TIME = 300;
```
781calmstate, line 21: since you already declared the default debounce time const, does it make sense to use it as a default argument in your `debounce` function instead of magic number (300)

src/App.tsx
```diff
@@ -1,16 +1,88 @@
...
+    const entry = input.trim().toLowerCase();
+
+    if (selectedPerson) {
+      setSelectedPerson(undefined);
+    }
+
+    if (entry !== appliedQuery) {
+      applyQuery(entry);
+    }
+  };
```
781calmstate, line 59: in react it's a good practice to use certain naming when you creating event handler functions.
the even handlers should be named with prefix `handle-`. The `updateQuery` name is also not very correct naming in your case, since in your function you are not only updating the query, right?
I suggest renaming the function into `handleChange` / `handleInputChange` / `handleSearchChange`, you got the idea.

Important mention, you name event handlers with `handle-` prefix, but the in the props they should be name with `on-` prefix - `onChange`, `onClick` etc.

```
const Component = () => {
   const handleChange = () => ....;

   return <Component2 onChange={handleChange} />
}

const Component2 = ({ onChange }) => {...}
 ```

Comment (781calmstate) [reply src/App.tsx:59]: You can read more about react namings conventions[ here](https://dev.to/kristiyanvelkov/react-js-naming-convention-lcg)

File: src/App.tsx [hunk 3]
```diff
@@ -1,16 +1,88 @@
...
+
+      return;
+    }
+
+    setFilteredPeople(
+      peopleFromServer.filter((person: Person) =>
+        person.name.trim().toLowerCase().includes(appliedQuery),
+      ),
+    );
+  }, [appliedQuery]);
```
781calmstate, line 47: my personal recommendation is to keep the `useEffect` somewhere on the bottom of the component.
some developers consider it also a good practice, but it's optional. Just easier to read imho.

I usually am trying to stick to something similar to this:

1. useStates, useRefs etc.
2. consts
3. handlers
4. useEffects

src/App.tsx
```diff
@@ -1,16 +1,88 @@
...
 
   return (
     <div className="container">
       <main className="section is-flex is-flex-direction-column">
-        <h1 className="title" data-cy="title">
-          {`${name} (${born} - ${died})`}
-        </h1>
+        {selectedPerson ? (
+          <h1 className="title" data-cy="title">
+            {`${selectedPerson.name} (${selectedPerson.born} - ${selectedPerson.died})`}
```
781calmstate, line 79: perhaps it's a good idea to just move it into the const for better readability

src/App.tsx
```diff
@@ -19,55 +91,35 @@ export const App: React.FC = () => {
               placeholder="Enter a part of the name"
               className="input"
               data-cy="search-input"
+              onChange={ChangeEvent => updateQuery(ChangeEvent.target.value)}
```
781calmstate, line 94: why `(e) => updateQuery(e.target.value)` doesn't work for you?

781calmstate (changes requested):
Your demo link isn't working.

As far as I can see your useCallbacks and useEffects implemented just fine, i didn't noticed any error in your implementation

PR #1050, opened 2025-10-09
_{create autocomplete}_

S-Khraban:
[DEMO LINK](https://S-Khraban.github.io/react_autocomplete/)


src/Autocomplete/Autocomplete.tsx
```diff
@@ -0,0 +1,124 @@
...
+      lastFilteredRef.current = '';
+    }
+  };
+
+  const handleBlur = () => {
+    window.setTimeout(() => setIsOpen(false), 100);
+  };
+
+  const handleChange = (e: React.ChangeEvent<HTMLInputElement>) => {
+    const next = e.target.value;
```
781calmstate: what does the name "next" suppose to mean? I can see you setting it to the query state which represents the search value I suppose. And then we have const named `next`, why?

src/Autocomplete/Autocomplete.tsx
```diff
@@ -0,0 +1,124 @@
...
+  onSelected,
+  delayMs = 300,
+  placeholder = 'Enter a part of the name',
+}) => {
+  const [query, setQuery] = useState<string>('');
+  const [isOpen, setIsOpen] = useState<boolean>(false);
+  const [suggestions, setSuggestions] = useState<Person[]>([]);
+
+  const debounced = useDebounce(query, delayMs);
+  const lastFilteredRef = useRef<string>('');
```
781calmstate: as I can see, you store the search query in this `lastFilteredRef`, if so then I don't understand why `lastFiltered` is the right name?

- `lastSearchQueryRef`
- `lastSearchRef`
- `lastQueryRef`
- `prevSearchRef`
All of them are giving us a hint that this ref is storing the search query value instead of something last filtered

src/Autocomplete/Autocomplete.tsx
```diff
@@ -0,0 +1,124 @@
...
+    );
+
+    setSuggestions(result);
+  }, [debounced, isOpen, people]);
+
+  const showNoMatches = useMemo(() => {
+    const typed = debounced.trim();
+
+    return isOpen && typed !== '' && suggestions.length === 0;
+  }, [debounced, isOpen, suggestions.length]);
```
781calmstate: this is a `boolean` value, which is a `primitive`, we don't need `useMemo` for `primitives`.

`useMemo` letting us to cache the value **to prevent unnecessary re-renders or prevent heavy re-computations**. 
What sense does it have to cache a `primitive`? for example we cache a `boolean` - it can be either true or false. Why would we want to cache a value that can be either one or another primitive + it's not a heavy computation which makes `useMemo` redundant here.

Sometimes, it could make sense to "cache" the non-primitive values, like objects or arrays because as you can remember **the variable stores the reference to the object instead of object itself**. **It creates new reference for object every time it re-renders** - which sometimes can force us to memoize it to store the stable reference unless we need to change the value.

src/Autocomplete/Autocomplete.tsx
```diff
@@ -0,0 +1,124 @@
...
+
+  return (
+    <>
+      <div className={`dropdown ${isOpen ? 'is-active' : ''}`}>
```
781calmstate: classnames

src/App.tsx
```diff
@@ -1,73 +1,29 @@
...
 import './App.scss';
-import { peopleFromServer } from './data/people';
+
+import { Person } from './types/Person';
+import { peopleFromServer as people } from './data/people';
+import { Autocomplete } from './Autocomplete/Autocomplete';
 
 export const App: React.FC = () => {
-  const { name, born, died } = peopleFromServer[0];
+  const [selected, setSelected] = useState<Person | null>(null);
```
781calmstate: selected what? It's selected person, so let's name it this way to make the life for other developer(your potential teammate) a bit easier

781calmstate (changes requested):
Pay more attention to the variables namings, they should be clear and descriptive. Other developer should be able to understand the purpose of the variable without need to figure out your code.

Start using the `classnames` 😁

mkyloa (approved):

PR #1096, opened 2026-01-12
Solution

DemonWayne:
[DEMO LINK](https://demonwayne.github.io/react_autocomplete/)


Dimononon (approved):
Good Job! One little remark: when the user stops focusing the input without selecting a person, dropdown list still renders, which is strange. Maybe add onBlur() for input to hide list when not needed?

PR #1142, opened 2026-03-29
solution

danf4691:
[DEMO LINK](https://danf4691.github.io/react_autocomplete/)


IanLuan (changes requested):
Great progress! However, there are two issues with your task:

1. When focusing on the empty input, the list is displayed, but we cannot select any items.
2. The debounce is not working properly. The search is triggering on every keystroke, but it should be debounced.

danf4691:
Closed the pull on mistake, the task is done now

IanLuan (approved):

PR #1172, opened 2026-05-05
Solution

OlehYuriev:
[DEMO LINK](https://OlehYuriev.github.io/react_autocomplete/)


src/components/Dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,40 @@
+import { memo } from 'react';
+import { Person } from '../../types/Person';
+
+type Prop = {
+  visible: boolean;
```
781calmstate, line 5: it's common practice to name boolean consts, props, arguments  using prefixes like `has`, `is`, `are`. e.g. `hasAccess`, `areEquals` etc

src/components/Dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,40 @@
...
+              <button
+                type="button"
+                role="menuitem"
+                key={item.slug}
+                className="dropdown-item"
+                data-cy="suggestion-item"
+                onClick={() => onSelected(item)}
+              >
+                <p
+                  className={`${item.sex === 'f' ? 'has-text-danger' : 'has-text-link'}`}
```
781calmstate, line 29: use `classnames` for conditional classes

src/App.tsx
```diff
@@ -1,73 +1,113 @@
...
+  useRef,
+  useState,
+} from 'react';
 import './App.scss';
 import { peopleFromServer } from './data/people';
+import { Person } from './types/Person';
+import { Dropdown } from './components/Dropdown/Dropdown';
+import { useDebounce } from './hooks/useDebounce';
+
+function filteredPerson(people: Person[], query: string): Person[] {
```
781calmstate, line 14: bad function naming, function names usually should indicate an action. in your case it's named like it is a filtered person and not an action

src/App.tsx
```diff
@@ -1,73 +1,113 @@
...
+  const queryLowerCase = query.trim().toLowerCase();
+
+  if (query !== '') {
```
781calmstate, line 17: i assume the check should be `queryLowerCase !== ''` instead. 

in your current functionality you are trimming the `queryLowerCase` but checking `query` which means that if `query` has empty spaces like `' '` it would be valid for if condition

src/components/Dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,40 @@
+import { memo } from 'react';
+import { Person } from '../../types/Person';
+
+type Prop = {
+  visible: boolean;
+  people: Person[];
+  onSelected: (person: Person) => void;
+};
+export const Dropdown = memo(function Dropdown({
```
781calmstate, line 9: it looks like unnecessary memoization

src/App.tsx
```diff
@@ -1,73 +1,113 @@
...
+        <div className="dropdown is-active" ref={refDropDown}>
           <div className="dropdown-trigger">
             <input
+              value={search}
+              onChange={e => {
+                setSearch(e.target.value);
+                if (selectedPerson) {
+                  setSelectedPerson(null);
+                }
+              }}
```
781calmstate, line 82: inline event handlers that exceeds one line should be moved into named handlers

781calmstate (changes requested):

